PR #26, opened 2024-06-14
Fix panic when WM_NAME has invalid UTF-8

galister:
I encountered this panic when running an old windows program via wine.

Using `from_utf8_lossy` instead will replace invalid characters with �

I also changed NET_WM_NAME, which as we know supposed to be valid UTF-8, but I believe it's better to not panic, just in case it isn't.


Supreeeme:
Thanks!

Issue #25, opened 2024-06-12
Panic: invalid HopSlotMap key used


galister:
Encountered this crash while dragging textures into texture slots in the Unity 2022.3.6f1 material editor.

```
thread 'main' panicked at src/server/mod.rs:542:53:
invalid HopSlotMap key used
```

Logs and backtrace: https://gist.github.com/galister/9db7b67df444d14f25257ef11507cff1

PR #29, opened 2024-06-21
fix crash when handling events with stale HopSlotMap key

galister:
all this does is switch the HopSlotMap access from indexed to checked, except in places where we handle the creation of a window, etc.

fixes crashes with Unity.

fixes https://github.com/Supreeeme/xwayland-satellite/issues/25

### DIFF
--- a/src/server/mod.rs
+++ b/src/server/mod.rs
@@ -496,13 +496,13 @@ impl<C: XConnection> ServerState<C> {
         handle_globals::<C>(&self.dh, globals.iter());
     }
 
-    fn get_object_from_client_object<T, P: Proxy>(&self, proxy: &P) -> &T
+    fn get_object_from_client_object<T, P: Proxy>(&self, proxy: &P) -> Option<&T>
     where
         for<'a> &'a T: TryFrom<&'a Object, Error = String>,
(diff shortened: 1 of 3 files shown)


src/server/mod.rs
```diff
@@ -568,14 +576,18 @@ impl<C: XConnection> ServerState<C> {
...
+                    let surface: &SurfaceData = object.as_ref();
+                    if let Some(SurfaceRole::Toplevel(Some(data))) = &surface.role {
+                        if let Some(min_size) = &hints.min_size {
+                            data.toplevel.set_min_size(min_size.width, min_size.height);
+                        }
+                        if let Some(max_size) = &hints.max_size {
+                            data.toplevel.set_max_size(max_size.width, max_size.height);
+                        }
+                    } else {
+                        warn!("Could not set size hint on {window:?}: stale surface")
```
Supreeeme, line 590: Isn't this else at the wrong level? I think it'll even warn for size hints on popups.

galister: fixed

Supreeeme:
Sorry for the delay, I forgor 💀 
Thanks!

PR #30, opened 2024-06-21
clipboard-sync behind feature

galister:
this is so that people who don't want implicit clipboard sync can just disable it.


Supreeeme:
I can't imagine why disabling clipboard syncing would ever be desirable. If it doesn't work properly it should just be fixed.

galister:
I agree, but it's clearly in an experimental state right now, with no easy fix for users that can't code (depending on the app, older commits crash on other errors, newer commits crash on clipboard events).

PR #37, opened 2024-07-06
Systemd notification

my4ng:
Currently when `xwayland-satellite` is run from a systemd service, there is no method to notify that initialisation has been completed for services that depend on it. Service types such as `simple` and `exec` will consider the service started right after forking (and `execve`) and dependent services may be started prematurely.

This PR adds a `systemd` feature that notifies systemd (with the `notify` type) when the setup is complete the dependent services may be started.


Supreeeme:
Nice, I meant to do this at some point but never got around to it.  Thanks!

PR #53, opened 2024-09-13
add: nix flake

Gavin-Niederman:
This adds a nix flake so that people who use nix can easily build it by just running ``nix develop``, ``nix build``, or ``nix run`` and so that it can easily be installed on nixos.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -50,16 +50,16 @@
     },
     "nixpkgs_2": {
       "locked": {
-        "lastModified": 1725983898,
-        "narHash": "sha256-4b3A9zPpxAxLnkF9MawJNHDtOOl6ruL0r6Og1TEDGCE=",
+        "lastModified": 1725930920,
+        "narHash": "sha256-RVhD9hnlTT2nJzPHlAqrWqCkA7T6CYrP41IoVRkciZM=",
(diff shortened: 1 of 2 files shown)


Supreeeme:
I don't actually use nix myself so I can't guarantee this will be kept up to date, but I tested it out and it seems to work well. Thanks!

sodiboo:
> I can't guarantee this will be kept up to date

It's worth noting that [i am maintaining up-to-date packages as part of my niri-flake](https://github.com/sodiboo/niri-flake/blob/de987fee96ca9ab53dcd61f0c39ab8efdd6066cb/flake.nix#L168-L214), with a packaging setup symmetric to niri ([stable and unstable](https://github.com/sodiboo/niri-flake/blob/de987fee96ca9ab53dcd61f0c39ab8efdd6066cb/flake.nix#L16-L20)). Currently i'm doing nothing more in terms of modules to make it easier to install, but the packages are present. If those suddenly break, i'll keep in mind the upstream flake and try to fix it if i have time. Also note that my niri-flake *isn't* redundant with adding a flake to this repository, because the flake in this repo is still useful for any NixOS users wanting to develop and contribute to xwayland-satellite

PR #61, opened 2024-10-17
Tablet support

Supreeeme:


Gavin-Niederman:
As soon as I bring the pen in range of the tablet, xwayland-satellite crashes.
Error message:
```
thread 'main' panicked at /build/source/src/server/event.rs:954:69:
called `Option::unwrap()` on a `None` value
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

vgskye:
also seeing above issue. fixed by following patch:
```patch
diff --git a/src/server/event.rs b/src/server/event.rs
index 8e6de9f..d9929eb 100644
--- a/src/server/event.rs
+++ b/src/server/event.rs
@@ -2,7 +2,7 @@ use super::*;
 use crate::clientside::LateInitObjectKey;
 use log::{debug, trace, warn};
 use macros::simple_event_shunt;
-use std::collections::HashSet;
+use std::{collections::HashSet, ops::Deref};
 use std::os::fd::AsFd;
 use wayland_client::{protocol as client, Proxy};
 use wayland_protocols::{
@@ -951,7 +951,7 @@ impl HandleEvent for TabletTool {
                 ProximityIn {
                     serial,
                     |tablet| {
-                        let key: ObjectKey = tablet.data().copied().unwrap();
+                        let key: ObjectKey = tablet.data::<LateInitObjectKey<TabletClient>>().map(Deref::deref).copied().unwrap();
                         let Some(tablet): Option<&Tablet> = state.objects.get(key).map(|o| o.as_ref()) else {
                             return;
                         };
```

Supreeeme:
Oopsie. Should be fixed, try again

Gavin-Niederman:
I'm still getting the same error as before, but if I use @vgskye's patch everything works as intended.

Supreeeme:
Wow, there were actually two lines like that lol. Hopefully got it this time.

Gavin-Niederman:
As far as I can tell it's all working! Thanks for implementing this!

vgskye:
seems to be working here as well, thanks!

Supreeeme:
Added a smoke test. Will merge as long as it still works. Mind doing a final sanity test? @Gavin-Niederman @vgskye

vgskye:
seems to work!
image

Gavin-Niederman:
It works! 
image

PR #63, opened 2024-10-18
Add optional systemd support to flake package

e-tho:
I've made systemd support optional in the flake package and enabled it by default as NixOS relies on it. Therefore, I've added an `xwayland-satellite-nosd` derivation in case anyone uses this flake with a Nix setup on distributions with different init systems.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -45,8 +45,13 @@
               xorg.libxcb
 
               makeWrapper
+
+              systemd
             ];
+
             buildInputs = [ pkgs.xwayland ];
+
+            cargoBuildOptions = opts: opts ++ [ "--features" "systemd" ];
 
             postInstall = ''
               wrapProgram $out/bin/xwayland-satellite \
--- a/flake.nix
+++ b/flake.nix
@@ -15,11 +15,39 @@
       let
         overlays = [ (import rust-overlay) ];
         pkgs = import nixpkgs { inherit system overlays; };
+        lib = pkgs.lib;
+
         naersk' = pkgs.callPackage naersk {
           cargo = pkgs.rust-bin.stable.latest.default;
           rustc = pkgs.rust-bin.stable.latest.default;
         };
-      in {
+
+        buildXwaylandSatellite = { withSystemd ? false }: naersk'.buildPackage {
+          src = ./.;
+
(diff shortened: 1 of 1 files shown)


Supreeeme:
Seems to work. Thanks!

flake.nix
```diff
@@ -33,26 +61,9 @@
...
-              makeWrapper
-            ];
-            buildInputs = [ pkgs.xwayland ];
-
-            postInstall = ''
-              wrapProgram $out/bin/xwayland-satellite \
-                --prefix PATH : ${pkgs.xwayland}/bin
-            '';
-          };
+          xwayland-satellite-nosd = buildXwaylandSatellite { withSystemd = false; };
```
sodiboo, line 64: Why is this a separate package output, rather than using `pkgs.callPackage` to make it overridable [like is done in nixpkgs](https://github.com/NixOS/nixpkgs/blob/d084336c705fb55a6c0daa247adb2d99d091810c/pkgs/by-name/xw/xwayland-satellite/package.nix#L11)? That way, `packages.<system>.xwayland-satellite-nosd` would instead be spelled `packages.<system>.xwayland-satellite.override { withSystemd = false; }`.

flake.nix
```diff
@@ -15,11 +15,39 @@
...
+          nativeBuildInputs = with pkgs; [
+            rustPlatform.bindgenHook
+            rust-bin.stable.latest.default
+            pkg-config
+
+            xcb-util-cursor
+            xorg.libxcb
+
+            makeWrapper
+          ] ++ lib.optional withSystemd pkgs.systemd;
```
sodiboo, line 37: Why is `systemd` added as a build input? `xwayland-satellite` only uses the `sd_notify` crate, which doesn't have any native dependency. [It writes to a text-based socket that systemd provides](https://github.com/lnicola/sd-notify/blob/6990e3733f69e5b23dad8985d66213dfb31675b6/src/lib.rs#L148-L151) and does little else; this dependency is pointless (and for instance, nixpkgs does not have it)

flake.nix
e-tho: > I think nixpkgs is downstream from us.

Good point, you got it. 

I've opened a PR (#65) addressing the mentioned changes, feel free to review it.

PR #65, opened 2024-10-23
Refactor flake for overridability, conditional service install, version formatting

e-tho:
This updates the flake to make the package overridable and install the service conditionally based on the `withSystemd` option. It introduces dynamic version formatting as `cargoPackageVersion-unstable-shortCommitHash` and cleans up unnecessary variables and inputs.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -52,7 +52,7 @@
               ${lib.optionalString withSystemd ''
                 install -Dm0644 resources/xwayland-satellite.service -t $out/lib/systemd/user
                 substituteInPlace $out/lib/systemd/user/xwayland-satellite.service \
-                  --replace '/usr/local/bin/xwayland-satellite' "$out/bin/xwayland-satellite"
+                  --replace-fail '/usr/local/bin/xwayland-satellite' "$out/bin/xwayland-satellite"
               ''}
               wrapProgram $out/bin/xwayland-satellite \
                 --prefix PATH : "${lib.makeBinPath [ pkgs.xwayland ]}"
--- a/flake.nix
+++ b/flake.nix
@@ -82,15 +82,7 @@
             };
           };
 
-        xwayland-satellite = pkgs.callPackage buildXwaylandSatellite {
-          lib = pkgs.lib;
-          rustPlatform = pkgs.rustPlatform;
-          pkg-config = pkgs.pkg-config;
-          makeBinaryWrapper = pkgs.makeBinaryWrapper;
-          libxcb = pkgs.xorg.libxcb;
-          xcb-util-cursor = pkgs.xcb-util-cursor;
-          xwayland = pkgs.xwayland;
-        };
+        xwayland-satellite = pkgs.callPackage buildXwaylandSatellite { };
       in
       {
         devShell = (pkgs.mkShell.override { stdenv = pkgs.clangStdenv; }) {
--- a/flake.nix
+++ b/flake.nix
@@ -23,8 +23,17 @@
 
         version = "${cargoPackageVersion}-${commitHash}";
 
-        buildXwaylandSatellite = { withSystemd ? true, ... } @ args:
-          pkgs.rustPlatform.buildRustPackage (rec {
+        buildXwaylandSatellite =
+          { withSystemd ? true
+          , rustPlatform
+          , pkg-config
+          , makeWrapper
+          , libxcb
+          , xcbutilcursor
+          , xwayland
+          }:
+
+          rustPlatform.buildRustPackage rec {
             pname = "xwayland-satellite";
             inherit version;
 
@@ -35,17 +44,19 @@
               allowBuiltinFetchGit = true;
             };
 
-            nativeBuildInputs = with pkgs; [
+            nativeBuildInputs = [
               rustPlatform.bindgenHook
               pkg-config
               makeWrapper
             ];
 
-            buildInputs = with pkgs; [
-              xorg.libxcb
-              xorg.xcbutilcursor
+            buildInputs = [
+              libxcb
+              xcbutilcursor
             ];
-           buildNoDefaultFeatures = true;
+
+            buildNoDefaultFeatures = true;
+
             buildFeatures = lib.optionals withSystemd [ "systemd" ];
 
             postInstall = ''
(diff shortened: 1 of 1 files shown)


sodiboo (changes requested):
it would be a lot nicer to read individual commit diffs, especially with e.g. switching out `naersk` for `rustPlatform` which makes the total diff somewhat painful to look at because every line changed. You probably made all of these changes before committing anything, and that's fair enough, but pretty please in general try to separate refactor commits (large diffs that basically make no change in functionality) from other types of changes?

e-tho:
> it would be a lot nicer to read individual commit diffs

I pushed the changes quickly due to time constraints after refactoring. I understand it's not ideal, but I figured it was acceptable since it only affects the flake and not the core project.

flake.nix
```diff
@@ -60,11 +83,8 @@
           ];
         };
 
-        packages = rec {
-          xwayland-satellite-nosd = buildXwaylandSatellite { withSystemd = false; };
```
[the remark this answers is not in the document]
e-tho: I understand the concern, but the number of people using this flake with a Nix installation on a non-systemd distribution is probably already close to zero, if not zero. Considering this was only merged yesterday, I think we can safely skip adding a warning and be confident that no one will be affected by this.

flake.nix
```diff
@@ -6,46 +6,69 @@
...
 
-        buildXwaylandSatellite = { withSystemd ? false }: naersk'.buildPackage {
-          src = ./.;
+        commitHash = self.shortRev or self.dirtyShortRev or "unknown";
 
-          nativeBuildInputs = with pkgs; [
-            rustPlatform.bindgenHook
-            rust-bin.stable.latest.default
-            pkg-config
+        version = "${cargoPackageVersion}-unstable-${commitHash}";
```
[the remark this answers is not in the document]
e-tho: > Isn't this kind off wrong given that at some point, a commit will be tagged for release, and then suddenly the release tag's package is called unstable? i think assigning an unstable label in this repo feels weird.

I think you're right about making the change, but not for the reason you mentioned. Whether or not the commit is tagged for release, using 'unstable' would have provided a human-readable way to indicate it's built from the main branch. However, since the user controls the input URL and can select a specific release, branch, or commit, 'unstable' wouldn't make sense in those cases.

> it also works weirdly given that the build output is now dependent on the git state

That's a valid point, but I don't find it very restrictive. I'm perfectly ok with that.

flake.nix
```diff
@@ -6,46 +6,69 @@
...
 
-            xcb-util-cursor
-            xorg.libxcb
+        buildXwaylandSatellite = { withSystemd ? true, ... } @ args:
```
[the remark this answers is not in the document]
e-tho: Indeed, I had only made the `withSystemd` option overridable because that was the primary goal of the refactoring without considering the dependencies. This is now addressed in 4c74de261f22165b1ea7002380ed55530b50a8b4.

Supreeeme:
If it looks good to the other nix people it looks good to me.
@sodiboo @Gavin-Niederman

flake.nix
```diff
@@ -6,46 +6,91 @@
...
+          };
+
+        xwayland-satellite = pkgs.callPackage buildXwaylandSatellite {
+          lib = pkgs.lib;
+          rustPlatform = pkgs.rustPlatform;
+          pkg-config = pkgs.pkg-config;
+          makeBinaryWrapper = pkgs.makeBinaryWrapper;
+          libxcb = pkgs.xorg.libxcb;
+          xcb-util-cursor = pkgs.xcb-util-cursor;
+          xwayland = pkgs.xwayland;
```
sodiboo, line 92: `pkgs.callPackage` prepopulates this. that's the whole point of `pkgs.callPackage`; it's `makeOverridable` but with any attribute from nixpkgs already filled in.

```suggestion
```

`libxcb` won't be filled in automatically though, i think. in that case, the closure should take `xorg` instead, and internally use `xorg.libxcb`.

also, is `libxcb` necessary? other packages (niri-flake, nixpkgs) don't mention it. is some functionality broken there as a result? cc @Supreeeme

sodiboo: other than being more concise, there is a subtle difference with `callPackage`: it will pass the correct versions of dependencies for *cross-compilation*. i guess we don't expose cross-comp here? to make use of this, the `pkgs` in `pkgs.callPackage` must be made for cross-compilation. and here, the `pkgs` definitely isn't cross-compilable. but in general, when the buiild and target platform are not identical, then manually specifying the dependencies is actually going to cause problems because `pkgs.*` is for the target platform, but e.g. `rustPlatform` should come from the build platform.

e-tho: > `pkgs.callPackage` prepopulates this.

Addressed in 8c56440.

> other than being more concise, there is a subtle difference with `callPackage`: it will pass the correct versions of dependencies for _cross-compilation_.

Got it.

> also, is `libxcb` necessary? other packages (niri-flake, nixpkgs) don't mention it.

`libxcb` is indeed included as a dependency in `xwayland-satellite`'s nixpkgs [expression](https://github.com/NixOS/nixpkgs/blob/7eee17a8a5868ecf596bbb8c8beb527253ea8f4d/pkgs/by-name/xw/xwayland-satellite/package.nix#L32).

Supreeeme: re: libxcb, the library used for the x11 protocol is in fact a rust wrapper around libxcb 🙂

sodiboo: > `libxcb` is indeed included as a dependency in `xwayland-satellite`'s nixpkgs [expression](https://github.com/NixOS/nixpkgs/blob/7eee17a8a5868ecf596bbb8c8beb527253ea8f4d/pkgs/by-name/xw/xwayland-satellite/package.nix#L32).

oh. i missed that. [i don't have it in `niri-flake`](https://github.com/sodiboo/niri-flake/blob/d7c058c0c0b8cd5b799a3a3ac4f8a1090454155e/flake.nix#L228-L229), is that package broken as a result then?

but yeah, seems i was wrong. this is probably necessary.

Supreeeme:
Assuming @sodiboo is satisfied with the current state, i'll merge this by the end of the day (about 10 hours from now)

sodiboo (approved):
yea. sounds good to me

PR #79, opened 2024-12-06
fix: add Event::Move check preventing infinite recursion caused by stale surfaces

### LINKED ISSUE #74: `Event::Motion` entering a stale surface overflows stack
## Description

In `Pointer::handle_event` on the `client::wl_pointer::Event::Motion` branch, if the pointer is pending enter, an `Event::Enter` and an `Event::Motion` to be handled on the new surface are added to a list which calls `handle_event` recursively. If the `Event::Enter` event finds a stale surface, the first call breaks early, but the second event, being identical to the event which spawned it, creates more recursive calls. Since the first event will still be stale, the second event keeps recursively calling itself until the program stack overflows, leaving behind a slew of "`could not enter surface: stale surface`" warnings.

## Notes

* I do not have a consistent replication method for this bug, though running [chatterino7](https://github.com/SevenTV/chatterino7) has given me a handful of these scenarios well after successful startup.
* I am using Niri, which might require dealing with the caveat mentioned in the comment, though I believe I understand the code well enough to say that this implementation is flawed regardless of compositor.

En-En-Code:
Fixes #74 (plus its duplicates)
Adds a check to Event::Move to check if the surface it would attempt to enter and stale and prevent recursive Event::Enter and Event::Move from taking place if it would.
Also did a small refactor to elimiante `process_event`, with the new method to make the recursion more apparent.
I will not claim the solution to be optimal, but I am confident it works armed with a reproduction method for my original bug.

### DIFF
--- a/src/server/event.rs
+++ b/src/server/event.rs
@@ -346,7 +346,6 @@ impl HandleEvent for Pointer {
         // destroy the menu if this occurs within a 500 ms interval (which it always does with
         // Niri). Other compositors do not run into this problem because they appear to not send
         // wl_pointer.enter until the user actually moves the mouse in the popup.
-        let mut process_event = Vec::new();
         match event {
             client::wl_pointer::Event::Enter {
                 serial,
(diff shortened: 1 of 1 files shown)


Supreeeme:
Thanks a lot for the fix! It looks much better than whatever I had going on before 🙂
Can you confirm `cargo test` passes, and can you be sure to run `cargo fmt`?

En-En-Code:
- [x] `cargo test` passes all tests (there should probably be a CI for that)
- [x] `cargo fmt` gets run by my editor on every save, but double-checked that there were no formatting differences

Supreeeme:
I'll set up CI eventually :) thanks for the fix!

PR #115, opened 2025-03-08
Support XDG Activation

bbb651:
This doesn't have support for clients doing [XSetInputFocus](https://tronche.com/gui/x/xlib/input/XSetInputFocus.html), I tried to support them with [Input Focus Events](https://tronche.com/gui/x/xlib/events/input-focus/) but they seemed to be only fired for the root window when adding [`EventMask::FOCUS_CHANGE`](https://docs.rs/xcb/latest/xcb/x/struct.EventMask.html#associatedconstant.FOCUS_CHANGE)
I'm also unsure if unwrapping on getting the window pid is a good idea, afaict it's up to the server and doesn't need support from clients, so assuming a recent xorg version it should always work..?


src/clientside.rs
```diff
@@ -222,6 +225,7 @@ push_events!(XdgOutput);
 push_events!(WlTouch);
 push_events!(ZwpConfinedPointerV1);
 push_events!(ZwpLockedPointerV1);
+// push_events!(XdgActivationTokenV1);
```
Supreeeme, line 228:
```suggestion
```

src/lib.rs
```diff
@@ -1,6 +1,7 @@
 mod clientside;
 mod data_device;
 mod server;
+mod xdg_activation;
```
Supreeeme, line 4: This would be better as a module under `server`

src/server/mod.rs
```diff
@@ -805,6 +835,42 @@ impl<C: XConnection> ServerState<C> {
         }
     }
 
+    pub fn activate_window(&mut self, window: x::Window) {
+        let Some(activation_state) = self.activation_state.as_ref() else {
+            warn!("Cannot focus window because xdg activation is not supported: {window:?}");
```
Supreeeme, line 840: This warning is already covered by the first log that mentions xdg activation is missing and would probably be spammy so I think it's unnecessary.

src/server/mod.rs
```diff
@@ -805,6 +835,42 @@ impl<C: XConnection> ServerState<C> {
...
+            return;
+        };
+
+        let Some(last_focused_toplevel) = self.last_focused_toplevel else {
+            warn!("No last focused toplevel, cannot focus window {window:?}");
+            return;
+        };
+        let Some(win) = self.windows.get(&last_focused_toplevel) else {
+            warn!("Unkown last focused toplevel, cannot focus window {window:?}");
```
Supreeeme, line 849:
```suggestion
            warn!("Unknown last focused toplevel, cannot focus window {window:?}");
```

src/server/mod.rs
```diff
@@ -805,6 +835,42 @@ impl<C: XConnection> ServerState<C> {
...
     }
 
+    pub fn activate_window(&mut self, window: x::Window) {
+        let Some(activation_state) = self.activation_state.as_ref() else {
+            warn!("Cannot focus window because xdg activation is not supported: {window:?}");
+            return;
+        };
+
+        let Some(last_focused_toplevel) = self.last_focused_toplevel else {
+            warn!("No last focused toplevel, cannot focus window {window:?}");
```
Supreeeme, line 845: I think on these failure states, we should tell the xserver to focus either the last focused toplevel if available or nothing if not, so we avoid having the X and Wayland focus states out of sync.

src/server/mod.rs
```diff
@@ -923,6 +995,23 @@ impl<C: XConnection> ServerState<C> {
...
     }
 
+    fn handle_activations(&mut self) {
+        let Some(activation_state) = self.activation_state.as_ref() else {
+            return;
+        };
+        let globals = &mut self.clientside.globals;
+
+        for (window, token) in globals.pending_activations.drain(..) {
+            warn!("pending activation: {window:?} {token:?}");
```
Supreeeme, line 1005:
```suggestion
            debug!("pending activation: {window:?} {token:?}");
```

src/xstate/mod.rs
```diff
@@ -650,6 +661,18 @@ impl XState {
...
+    fn get_pid(&self, window: x::Window) -> u32 {
+        self.connection
+            .wait_for_reply(self.connection.send_request(&xcb::res::QueryClientIds {
+                specs: &[xcb::res::ClientIdSpec {
+                    client: window.resource_id(),
+                    mask: xcb::res::ClientIdMask::LOCAL_CLIENT_PID,
+                }],
+            }))
+            .map(|reply| reply.ids().next().unwrap().value()[0])
+            .expect("Failed to get pid of window")
```
Supreeeme, line 673: While I honestly wouldn't expect this to fail, I'd rather fail on the safe side and just return an Option and print a warning. I've ran into far too many things I wouldn't expect to fail...

src/xdg_activation.rs
```diff
@@ -0,0 +1,33 @@
...
+    activation::{ActivationHandler, RequestData, RequestDataExt},
+    delegate_activation,
+};
+use xcb::x;
+
+use crate::clientside::Globals;
+
+delegate_activation!(Globals, ActivationData);
+
+pub struct ActivationData(pub x::Window, pub RequestData);
```
Supreeeme, line 11: It seems like the members are pub only for the sake of construction, I'd rather they not be pub and have a `ActivationData::new` method.

Supreeeme (commented):
This could use a test or two. Be sure to run `cargo fmt`.

src/lib.rs
bbb651: Would this be in `clientside`..? I followed `data_device`, I guess they can also be under `clientside`, so the folder structure would be `clientside/{mod.rs data_device.rs, xdg_activation.rs}`?

src/server/mod.rs
bbb651: I'm not sure this is the right place to do it, since activation can fail and the protocol doesn't tell us if it did, and a desync also happens after XSetInputFocus. And I don't think it's much worse to have a different window get x11 focus rather than no window, the bigger problem is that currently in the state where a wayland window has wayland focus but no x11 focus (e.g. xterm after `xdotool search glxgears windowfocus`), it won't regain focus until you lose and regain wayland focus, I'll look into fixing this case.
It's hard to completely sync focus both ways, even if we can somehow watch for x11 input focus changes and try to revert it if activation fails (by not having a `wl_keyboard::leave` event), we cannot know how long we should wait before giving up.

src/lib.rs
Supreeeme: Yeah, that's fine too. Probably better.

src/server/mod.rs
Supreeeme: Hm, good point. We should probably generally watch for input focus and prevent them from getting out of sync, since like you said it can happen even without activation. But that should be separate from this PR.

Supreeeme:
As far as testing, it'll likely be simpler to write an integration test (under `tests/integration.rs`. You will have to add xdg activation support to testwl. I think it should be straightforward enough to copy the setup from other tests but if not feel free to ask.

bbb651:
I started working on the tests, I think I got the hang of it but this currently cannot work because the PID of the client is the same in all tests, so depending on the order the logic for new process activation through `/proc/{pid}/environ` will fail.
I think we need to either:
- Make things more generic over things that need IO to mock it in the test
- ~~Use something like https://nexte.st/ to run tests per-process rather than per-thread~~ Actually in this case this wouldn't help because `/proc/{pid}/environ` contains the initial environment
- [black magic](https://unix.stackexchange.com/a/438007)
- Only test in process window activation

Supreeeme:
Hm, I feel like the easiest path would be to just create another binary crate in the workspace and launch it from the test and add it to dev-dependencies (I think this is possible? I'm not actually sure). If you feel like that is an outsized amount of work, I would also accept only testing in process window activation, as I'm most interested in being sure the activation logic itself is working properly.

bbb651:
See the linked cargo issue, it's possible but it's a bit hacky, using a feature and a dev-dependency on yourself to activate it. Testing only in process activation seems good enough for now

bbb651:
Couldn't get the test to pass but I think it's a problem in the test, I don't think the `_NET_ACTIVE_WINDOW` is actually being sent properly, sleeping doesn't seem to help and trying to insert `f.wait_and_dispatch();` at any point results in a broken pipe...

The testwl side turned out a bit long but I'm pretty confident with it, I referenced niri and smithay and it should match niri's behavior pretty well, excluding the 10 second timeout because that either unnecessary or flaky for tests, there's no in-between.

Supreeeme (commented):
There have been some changes to the test framework that made the integration tests a little less flaky, so you should rebase. They're still somewhat flaky though so if you see pipes breaking just try rerunning it a couple times.

tests/integration.rs
```diff
@@ -725,6 +725,27 @@ fn input_focus() {
...
+    let window2 = connection.new_window(connection.root, 0, 0, 20, 20, false);
+    let surface2 = f.map_as_toplevel(&mut connection, window2);
+
+    f.testwl.focus_toplevel(surface1);
+    connection.set_property(
+        connection.root,
+        connection.atoms.net_active_window,
+        x::ATOM_WINDOW,
+        &[window2],
+    );
```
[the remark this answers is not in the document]
bbb651: Oh, I assumed ClientMessages were sent by the X server after certain actions like setting the `_NET_ACTIVE_WINDOW` property. Clients send client messages and the X server echos them? It doesn't look like I can even send them with `xcb`, I found nothing useful searching `ClientMessageEvent->` and `BaseEvent->`: https://docs.rs/xcb/latest/xcb/x/struct.ClientMessageEvent.html#impl-BaseEvent-for-ClientMessageEvent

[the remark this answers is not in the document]
bbb651: It worked! Just need to rebase now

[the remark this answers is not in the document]
bbb651: Ok I take it back, I was doing something wrong because it's not observable on the X client that the active window changed, I'm doing the same thing xdotool does...

bbb651:
Also I realized I completely forgot the case of a new x11 window (it happened to work when I tested it because wayland-aware launchers set `XDG_ACTIVATION_TOKEN` regardless).
I don't think we need to support `DESKTOP_STARTUP_ID`, desktop files could have `StartupNotify` and `StartupWMClass` but afaik it's just used by the desktop environment so it practically does nothing in this case.

tests/integration.rs
```diff
@@ -781,6 +792,59 @@ fn input_focus() {
...
+
+        assert_eq!(&[window2], reply.value::<x::Window>());
+    };
+
+    connection.send_client_message(&x::ClientMessageEvent::new(
+        window1,
+        connection.atoms.net_active_window,
+        x::ClientMessageData::Data32([2, x::CURRENT_TIME, 0, 0, 0]),
+    ));
+    std::thread::sleep(Duration::from_millis(50));
```
Supreeeme, line 827: Could this be `f.wait_and_dispatch()`?

testwl/src/lib.rs
```diff
@@ -266,9 +286,13 @@ impl State {
...
         }
     }
+    
+    fn get_focused(&self) -> Option<SurfaceId> {
+        self.keyboard.as_ref()?.current_focus
+    }
 
     #[track_caller]
-    fn configure_popup(&mut self, surface_id: SurfaceId) {
+    pub fn configure_popup(&mut self, surface_id: SurfaceId) {
```
Supreeeme, line 295: Don't think this needs to be pub

bbb651: I accidentally reverted it when rebasing https://github.com/Supreeeme/xwayland-satellite/commit/54a7ad9e13bdcd27fc30169877d7514a52097b0f

tests/integration.rs
bbb651: I tried that and it panics with `Did not receive any events`...
I also tried setting the flags and replacing `x::CURRENT_TIME` with `0` like [this code](https://github.com/PlugOvr-ai/PlugOvr/blob/bee8b73842602976711b9359138d242dbb5d4661/src/window_handling.rs#L44-L55), same result it's completely unobservable for the window manager

tests/integration.rs
```diff
@@ -546,6 +546,17 @@ impl Connection {
         })
         .unwrap();
     }
+
+    #[track_caller]
+    fn send_client_message(&self, request: &x::ClientMessageEvent) {
+        self.send_and_check_request(&x::SendEvent {
+            propagate: false,
+            destination: x::SendEventDest::Window(self.root),
+            event_mask: x::EventMask::empty(),
```
Supreeeme, line 555:
```suggestion
            event_mask: x::EventMask::SUBSTRUCTURE_REDIRECT | x::EventMask::SUBSTRUCTURE_NOTIFY
```

tests/integration.rs
```diff
@@ -781,6 +792,59 @@ fn input_focus() {
...
+            property: connection.atoms.net_active_window,
+            r#type: x::ATOM_WINDOW,
+            long_offset: 0,
+            long_length: 1,
+        });
+
+        assert_eq!(&[window2], reply.value::<x::Window>());
+    };
+
+    connection.send_client_message(&x::ClientMessageEvent::new(
```
Supreeeme, line 822:
```suggestion
    std::thread::sleep(Duration::from_millis(1));
    connection.send_client_message(&x::ClientMessageEvent::new(
```
For some reason, adding this mini sleep right before sending the client message enables it to actually get processed. Why? Don't know, maybe a race condition. Don't care too much.

testwl/src/lib.rs
```diff
@@ -1524,3 +1554,124 @@ impl Dispatch<WlCallback, ()> for State {
         unreachable!()
     }
 }
+
+impl GlobalDispatch<XdgActivationV1, ()> for State {
```
Supreeeme, line 1558: You need to call `dh.create_global` in `Server::new` - the activation global has not actually been exposed to Xwayland yet.

Supreeeme (commented):
Found the issues with the tests.

testwl/src/lib.rs
bbb651: It doesn't need to this happens entirely between `clientside` and the window manager, Xwayland never tries to bind xdg-activation

bbb651:
Something is still weird with the focus, and it needs:
```rs
    f.testwl.dispatch();
    std::thread::sleep(Duration::from_millis(10));
    f.testwl.dispatch();
    std::thread::sleep(Duration::from_millis(10));
    f.testwl.dispatch();
```
but it finally works, turns out I forgot to actually fire the token done event and `last_enter_serial` was never actually updated (and also I had my serial older than enter logic flipped, not that it actually matters because serials are incremented in all places) in testwl

tests/integration.rs
```diff
@@ -781,6 +792,33 @@ fn input_focus() {
...
+        window1,
+        connection.atoms.net_active_window,
+        x::ClientMessageData::Data32([2, x::CURRENT_TIME, 0, 0, 0]),
+    ));
+    f.testwl.dispatch();
+    std::thread::sleep(Duration::from_millis(10));
+    f.testwl.dispatch();
+    std::thread::sleep(Duration::from_millis(10));
+    f.testwl.dispatch();
```
Supreeeme, line 817: Looks like a `f.wait_and_dispatch()` candidate? :)

Supreeeme:
Looks great, thanks!

bbb651:
I noticed this doesn't work with flatpak, which is not very suprising, the `/proc/{pid}/environ` method is pretty fragile in general.
Niri actually also sets [`DESKTOP_STARTUP_ID`](https://github.com/YaLTeR/niri/blob/fd1f43673cf4d1e84d9da67e294c040934034eda/src/utils/spawning.rs#L132) as well, I'll look into implementing it.

PR #119, opened 2025-03-12
Add CI

Supreeeme:


En-En-Code:
Did the CI actually hit a stall on a test? :rofl: I've run these integration tests more than anyone recently and still haven't hit a stall outside of the initial one I reported. I feel kinda bad. Probably should figure out a way to fail if a test runs for too long.

Supreeeme:
Wow didn't realize it was still running lol. I suspect it's just the flakiness you reported, but yeah should definitely just fail tests  that take way too long

PR #120, opened 2025-03-13
Fix most Broken Pipe IO aborts

En-En-Code:
Makes sure the xwayland_satellite main thread ends so that the file descriptors don't continue to be used after Fixture drops them

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,84 @@
+name: CI
+
+on:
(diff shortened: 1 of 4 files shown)


Supreeeme:
Hope you don't mind, but I went ahead and pulled this into the CI branch.

En-En-Code:
As long as you don't mind me immediately making a "Refactors" PR. ;)

Issue #113, opened 2025-03-02
Unwrap on none in event.rs | client::wl_pointer::Event::Enter


ToxicMushroom:
Got a random crash today:
```
thread 'main' panicked at /xwayland-satellite/src/server/event.rs:381:70:
called `Option::unwrap()` on a `None` value
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

https://github.com/Supreeeme/xwayland-satellite/blob/44590a416d4a3e8220e19e29e0b6efe64a80315d/src/server/event.rs#L381C16-L381C79

PR #122, opened 2025-03-14
Late refactors

En-En-Code:


En-En-Code:
Welp, stalling was the one issue we did not make headway on. Thought they were between 1/50 to 1/75, but that immediately plunges to 0% when you run them in GDB, of course. (If you do run the integration binary in GDB, make sure to `handle SIGPIPE nostop` since the thread which reads `stderr` from `Xwayland` loves hitting `SIGPIPE` since the receiver closes first and that breakpoint messes all of the future tests up.

En-En-Code:
Since the tests stalled, I'll just also throw it the original reason for working on this again. I saw #113 , which I may have encountered once or twice, but never with a replication method, logs, or a backtrace, so I didn't report. Realized that the exact pattern showed up in a few other spots, so I took them out as well.

src/lib.rs
```diff
@@ -131,14 +133,18 @@ pub fn main(data: impl RunData) -> Option<()> {
 
     // `finish_rx` only writes the status code of `Xwayland` exiting, so it is reasonable to use as
     // the UnixStream of choice when not running the integration tests.
-    let quit_rx = data.quit_rx().unwrap_or(&finish_rx);
+    let (mut quit_rx, is_test_thread) = if let Some(quit_rx) = data.quit_rx() {
+        (quit_rx, true)
+    } else {
+        (finish_rx, false)
+    };
```
Supreeeme, line 140: Hm, what's the point of doing this?

Supreeeme (commented):
Could you make the first commit message more meaningful, and lose the merge commit?

src/lib.rs
En-En-Code: I was an idiot. Not necessary.

En-En-Code: I want to print an error if `quit_rx` has data only for the standalone program (since that is `finish_rx` which means `Xwayland` exited prematurely), but because `quit_rx` calls `take` on the underlying option (because it is in a mutex), it can't be checked at that later point.

En-En-Code:
I'm so sorry for whatever this abomination of a PR is. I might need my Git license revoked. (I've had like 5 meltdowns over the past few days doing fairly mundane tasks.)

src/lib.rs
Supreeeme: I have some mental reservations against putting explicit test checks in normal application code (although I acknowledge this is kind of hypocritical considering RunData pretty much explicitly exists for the sake of tests.) What I'd prefer is to have another method on the `RunData` trait, i.e. `fn on_xwayland_quit(status: Box<ExitStatus>)`, call `on_quit` in the loop when `quit_rx` gets some data, and have it do nothing in the test implementation and print the error for the regular application.

Supreeeme:
Looks good to me, thanks!

PR #124, opened 2025-03-17
Remove paste dependency

En-En-Code:
[RustSec Advisory - `paste` is unmaintained](https://rustsec.org/advisories/RUSTSEC-2024-0436.html)

Was not a large part of the total infrastructure, so I just wrote out the types paste would create myself as macro parameters. The macro syntax seems fine (macro syntax design is not something I would call myself *good* at though).


Supreeeme:
I don't see this as a particularly big deal, but paste is not used much anyway like you said so I'll accept some slightly lower compile times.

PR #134, opened 2025-03-24
Support xdg-decoration

bbb651:


src/server/mod.rs
```diff
@@ -1224,6 +1276,12 @@ impl<C: XConnection> ServerState<C> {
             toplevel.set_fullscreen(None);
         }
 
+        let decoration = self.decoration_manager.as_ref().map(|decoration_manager| {
+            let decoration = decoration_manager.get_toplevel_decoration(&toplevel, &self.qh, ());
+            decoration.set_mode(zxdg_toplevel_decoration_v1::Mode::ServerSide);
```
Supreeeme, line 1281: Shouldn't we check `win.attrs.decorations`?

Supreeeme (commented):
I wanted to also provide custom decorations (maybe using sctk-adwaita?) if the host compositor doesn't support server side decorations, but that doesn't have to be a part of this PR.

bbb651:
Yeah I think it can be a separate PR, might be a good opportunity to make it [reparenting](https://en.wikipedia.org/wiki/Re-parenting_window_manager) (I learned about this from here https://github.com/Smithay/smithay/issues/389 it's basically just to workaround wrong assumptions in applications, it's more annoying for us because we can't set the nonreparenting environment variable for java automatically) ~~and then use the frame window for CSD?~~ actually I think it's pointless and will just complicate keeping track of state and going through Xwayland

bbb651:
Oh also I forgot to mention but this entirely ignores double-buffering with commits and configure, I think at worst it means there'll be at most a couple of frames too large or small by the size of the decoration when the applications switches between decorations, haven't tested it outside niri actually, I'll test it in Louvre

Supreeeme:
Is this good to merge or should I wait?

bbb651:
Should be good to merge

PR #147, opened 2025-04-04
Set popup constraint adjustment to slide

bbb651:
Like @ehopperdietzel requested, I don't have a strong opinion on this but since afaik x11 doesn't have an equivalent for this (it barely has the concept of popups), it seems like a reasonable default.
Tested it a bit on gtk4 and ~~it works as expected~~ actually the part of the popup that overflows above the window doesn't get mouse events... it's not directly related to this but makes this pretty pointless, I'll try to investigate it

### DIFF
--- a/src/server/mod.rs
+++ b/src/server/mod.rs
@@ -26,6 +26,7 @@ use wayland_protocols::xdg::decoration::zv1::client::zxdg_decoration_manager_v1:
 use wayland_protocols::xdg::decoration::zv1::client::zxdg_toplevel_decoration_v1::{
     self, ZxdgToplevelDecorationV1,
 };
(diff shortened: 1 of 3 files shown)


Supreeeme:
If you can figure out the issue with SlideY I will merge. I recall adding this when I first implemented popups but ran into some issues that I can't recall, possibly could've been this. If I had to guess, this is related to the fact that satellite positions all windows at 0x0 from X's point of view, so a popup that goes above a window would have a negative Y coordinate and X does not like that. A solution might involve applying some kind of offset on the cursor when this kind of popup is detected.

Supreeeme:
This has been accomplished with 2dd40ecfe882c1659e0e46c761974702c7ab5fde.

Issue #31, opened 2024-06-25
Implement xdg-decorations


ehopperdietzel:
Hi, I just wanted to thank you for developing this project. It's such a clever solution! Do you plan to provide client-side decorations or implement the xdg-decoration protocol? That would be really nice since compositors would be able to know if windows are not decorated.

Fox2Code:
Note: https://github.com/CuarzoSoftware/Louvre

And yeah this project seems great!

Supreeeme:
Client side/xdg decorations are planned yes, just a bit lower on the priority list for now, I'd like to address some existing bugs first. But it shouldn't be too difficult to add.

bbb651:
I got I've the detection part working with `_MOTIF_WM_HINTS`:

<details>

<summary>motif-wm-hints.patch</summary>

```diff
diff --git a/src/xstate/mod.rs b/src/xstate/mod.rs
index 13143d1568..86fd5db024 100644
--- a/src/xstate/mod.rs
+++ b/src/xstate/mod.rs
@@ -8,7 +8,7 @@
 use std::ffi::CString;
 use std::os::fd::{AsRawFd, BorrowedFd};
 use std::rc::Rc;
-use xcb::{x, Xid, XidNew};
+use xcb::{x, Cookie, Xid, XidNew};
 use xcb_util_cursor::{Cursor, CursorContext};
 
 // Sometimes we'll get events on windows that have already been destroyed
@@ -491,6 +491,7 @@
         let class = self.get_wm_class(window);
         let wm_hints = self.get_wm_hints(window);
         let size_hints = self.get_wm_size_hints(window);
+        let motif_wm_hints = self.get_motif_wm_hints(window);
 
         let geometry = self.connection.wait_for_reply(geometry)?;
         debug!("{window:?} geometry: {geometry:?}");
@@ -503,6 +504,7 @@
         let class = class.resolve()?;
         let wm_hints = wm_hints.resolve()?;
         let size_hints = size_hints.resolve()?;
+        let motif_wm_hints = motif_wm_hints.resolve()?;
 
         Ok(WindowAttributes {
             override_redirect: attrs.override_redirect(),
@@ -517,6 +519,7 @@
             class,
             group: wm_hints.and_then(|h| h.window_group),
             size_hints,
+            decorations: dbg!(motif_wm_hints.and_then(|h| h.decorations)),
         })
     }
 
@@ -660,6 +663,28 @@
         }
     }
 
[...]

PR #148, opened 2025-04-04
Destory popups in FILO order

bbb651:
Fixes vscode issue described here: https://github.com/Supreeeme/xwayland-satellite/issues/31#issuecomment-2777257710

The logic in `Request::<WlSurface>::Destroy` is messy, I've had to wrestle with the borrow checker to get it to work (I upgraded to edition 2024 because I thought the I encountered a borrow checker limitation that was lifted but apparently not, it should probably be a separate PR).
A lot of tests are consistently failing, I have a feeling it's due to an existing race condition that is now a lot more likely because surface destruction now takes a more time.

Also I think storing the keys of children is overkill, using a count should be enough like in testwl. I think this simplifies the borrowing a bit.

Edit: turns out this was also the wrong place for it, the role is destroyed beforehand when the window is unmapped.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,9 +14,7 @@ all = "deny"
 [package]
 name = "xwayland-satellite"
 version = "0.5.1"
-edition = "2021"
-
-# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+edition = "2024"
 
 [lib]
 crate-type = ["lib"]
--- a/macros/Cargo.toml
+++ b/macros/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "macros"
 version = "0.1.0"
-edition = "2021"
+edition = "2024"
 
 [lints]
 workspace = true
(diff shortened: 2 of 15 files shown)


bbb651:
I tried switching to a refcount (children count + the role itself to avoid a child destroying the parent) which is better than the `HashSet` but it's still turning into a mess, the "zombie" popups are breaks assumptions in a lot of places. I also tried making `role: Option<Rc<RefCell<SurfaceRole>>>` it's a nightmare ergonomically especially for matching on the role, and now that I think about it doesn't even work because the client surface would be dropped before the role... @Supreeeme If I understand it correctly, windows are allowed to unmap then remap as a different role and the surface would be reused? If so we need to swap the `SurfaceData` for a new one and keep the old ones as zombie popups

Supreeeme:
I think a window could in theory switch roles, however in practice I have not ever seen this happen, so I wouldn't worry about it.

I also would try parenting all popups to a toplevel instead of trying to manage the destruction order. This would probably involve adding some sort of static offset to the popup positions.

Supreeeme:
I ended up having all popups parented to toplevels (which also fixed positioning issues), so this is not needed.

PR #149, opened 2025-04-05
Cursor shape

bbb651:
Mostly draft because disabling disabling the `client-cursors` feature causes weird behavior and crashes, we probably have to create our own fake cursor theme which we can then identify serials of.

It needs some testing (mainly with `client-cursors` enabled), I've made some very specific heuristics tailored to gtk for detecting default and hidden cursors, and also if there are any missing cursor name (they should print out as warnings).


Supreeeme:
I think an unset cursor should just be the default cursor.

Issue #102, opened 2025-02-01
On-demand/socket activation


YaLTeR:
Compositors like Mutter do a thing where they create and listen to a `DISPLAY` socket on their own, then only when a client connects, they spawn Xwayland on-demand and pass it the client fd. I'd like to do the same in niri, but with xwayland-satellite.

The advantages are:
1. niri locks/creates the socket, so it can export the right `DISPLAY` variable automatically.
2. If xwayland-satellite dies, the socket remains open on niri side, it can just spawn a new instance right away if a new client connects.
3. This also allows xwayland-satellite to exit when all X11 clients disconnect to conserve resources.

I'm not sure how exactly this works. Here's the relevant Mutter code: https://gitlab.gnome.org/GNOME/mutter/-/blob/fc6d79eda9395fe1151049b1b0a23974911bc1d7/src/wayland/meta-xwayland.c#L850-886 Seems like it spawns Xwayland with a few fds open, then tells it about them in commandline args. I guess this needs to be exposed in Smithay's Xwayland wrapper.

YaLTeR:
This is now implemented and works in git xwayland-satellite + git niri.

PR #169, opened 2025-06-03
Accept and pass -listenfd on to Xwayland

YaLTeR:
Implements the part of #102 necessary for compositor-driven on-demand activation. I've got a proof-of-concept working in niri using this, but I'll leave it as a draft until I flesh out the niri part some more.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -8,7 +8,7 @@ use log::{error, info};
 use rustix::event::{poll, PollFd, PollFlags};
 use smithay_client_toolkit::data_device_manager::WritePipe;
 use std::io::{BufRead, BufReader, Read, Write};
-use std::os::fd::{AsFd, AsRawFd, BorrowedFd};
+use std::os::fd::{AsFd, AsRawFd, BorrowedFd, OwnedFd};
 use std::os::unix::net::UnixStream;
 use std::process::{Command, ExitStatus, Stdio};
 use wayland_server::{Display, ListeningSocket};
@@ -35,6 +35,7 @@ type RealServerState = ServerState<RealConnection>;
 
 pub trait RunData {
     fn display(&self) -> Option<&str>;
+    fn listenfds(&mut self) -> Vec<OwnedFd>;
     fn server(&self) -> Option<UnixStream> {
         None
     }
@@ -46,7 +47,7 @@ pub trait RunData {
     fn xwayland_ready(&self, _display: String, _pid: u32) {}
 }
 
-pub fn main(data: impl RunData) -> Option<()> {
+pub fn main(mut data: impl RunData) -> Option<()> {
     let mut version = env!("VERGEN_GIT_DESCRIBE");
     if version == "VERGEN_IDEMPOTENT_OUTPUT" {
         version = env!("CARGO_PKG_VERSION");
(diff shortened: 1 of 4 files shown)


YaLTeR:
Alright seems to work. You can give it a try with https://github.com/YaLTeR/niri/pull/1728.

One thing is that I set the std file descriptors to NULL in niri bc I don't want xwl-s to spam the journal output. For debugging you can always spawn an xwl-s instance by hand with RUST_LOG and such.

Though, it might be nice to make xwl-s log to the journal too (I believe there's a tracing layer for that), then if I'm not mistaken, it should appear as separate journal entries even when spawned by niri with std descriptors closed. (fuzzel does this and it seems to work)

YaLTeR:
Also, I suspect it should now be trivial to add termination after inactivity (something along: pass `-terminate 10` to Xwayland and optionally don't panic when it dies), since niri already handles abnormal xwl-s termination by just continuing to watch the X11 socket.

src/lib.rs
```diff
@@ -35,6 +35,7 @@ type RealServerState = ServerState<RealConnection>;
 
 pub trait RunData {
     fn display(&self) -> Option<&str>;
+    fn listenfds(&self) -> &[RawFd];
```
Supreeeme, line 38: Maybe this could just return a Vec of `OwnedFd`, so we can just drop them and have them closed

src/main.rs
```diff
@@ -1,24 +1,58 @@
...
+        self.display.as_deref()
+    }
+
+    fn listenfds(&self) -> &[RawFd] {
+        &self.listenfds[..]
     }
 }
 
-fn get_display() -> Option<String> {
+fn parse_args(data: &mut RealData) {
```
Supreeeme, line 30: Maybe parse_args should just create and return a `RealData`.

src/main.rs
```diff
@@ -1,24 +1,58 @@
...
     let mut args: Vec<_> = std::env::args().collect();
-    if args.len() > 2 {
-        panic!("Unexpected arguments: {:?}", &args[2..]);
+    if args.len() < 2 {
+        return;
+    }
+
+    // Argument at index 1 is our display name. The rest can be -listenfd.
```
Supreeeme, line 36: I am not a big fan of 

1. requiring a display number when you want to pass `-listenfd` 
2. `-listenfd` having to come after the display number

I don't have a particularly good argument against 1 but I'd prefer for 2 to be resolved.

Supreeeme (commented):
Cool stuff!

src/main.rs
YaLTeR: 1 is how it works in [mutter](https://gitlab.gnome.org/GNOME/mutter/-/blob/48.3.1/src/wayland/meta-xwayland.c?ref_type=tags#L867) and [wlroots](https://gitlab.freedesktop.org/wlroots/wlroots/-/blob/0.19.0/xwayland/server.c?ref_type=tags#L48) so I did the same.

2 is just matching the argument order for Xwayland itself, there the display name also comes first. So pretty much yeah, I went for making it look like invoking Xwayland. If you'd like I can reorder the arguments, but then I'd also prefer to make them look more different from Xwayland's (e.g. `--listenfd` with two dashes). But I personally think that mirroring Xwayland's structure makes sense, especially when it's doing conceptually the same thing.

Supreeeme: Fair enough.

Issue #166, opened 2025-05-30
Discord & Spotify popups being windowed/tiled.


ToxicMushroom:
### Operating System

Arch linux

### Wayland Compositor

Niri-git 8ba57fc ~ [25.05.1](https://github.com/YaLTeR/niri/commit/8ba57fcf25d2fc9565131684a839d58703f1dae7)

### Describe the issue.

Popups are becoming windows.

![Image](https://github.com/user-attachments/assets/cfc4417f-43b9-4ad7-a6c2-4db9be8e3683)

![Image](https://github.com/user-attachments/assets/063a182f-b418-4510-a5df-76d7ea2a70ea)

I also have a similar issue in another gtk2 app where the type-ahead popups opens a window that instantly closes, breaking the type-ahead function in the file picker. 
But I think it might get fixed if this is fixed since it started happening at the same commit.

**Versions**
At the time of posting, the latest commit tested 3ba30b149f9eb2bbf42cf4758d2158ca8cceef73
Started happening at 4671f2728220fb555da878e39cdb12c667113b54
51300780f84cb61d04a3c469099c758461b2a816 works fine

**Debug Logs**
[discord-dnd-xwls-debug.log](https://github.com/user-attachments/files/20522728/discord-dnd-xwls-debug.log)

[spotify-dnd-xwls-debug.log](https://github.com/user-attachments/files/20522733/spotify-dnd-xwls-debug.log)

[verifast-typeahead-xwls-debug.log](https://github.com/user-attachments/files/20522734/verifast-typeahead-xwls-debug.log)

### Steps to reproduce

1. Open discord under x11, (dont pass ozone args)
2. Drag a server/channel/image around
3. A new window pops up, instead of a popup that follows the cursor

### Please confirm you have reproduced this on the latest commit.

- [x] I have reproduced this on the latest commit

PR #173, opened 2025-06-10
Recognize drag and drop window type

sky1e:
Recognize `_NET_WM_WINDOW_TYPE_DND` used by discord as a popup, fixing #166.

I tried to add an integration test for it, but found it was failing despite `guess_is_popup()` returning true for the test window. Ultimately I gave up trying to get the test to work correctly in service of submitting something, and marked the test `#[ignore]`.

### DIFF
--- a/tests/integration.rs
+++ b/tests/integration.rs
@@ -320,6 +320,7 @@ xcb::atoms_struct! {
         win_type_normal => b"_NET_WM_WINDOW_TYPE_NORMAL",
         win_type_menu => b"_NET_WM_WINDOW_TYPE_MENU",
         win_type_tooltip => b"_NET_WM_WINDOW_TYPE_TOOLTIP",
+        win_type_dnd => b"_NET_WM_WINDOW_TYPE_DND",
         motif_wm_hints => b"_MOTIF_WM_HINTS" only_if_exists = false,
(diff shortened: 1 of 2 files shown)


tests/integration.rs
```diff
@@ -1728,6 +1729,28 @@ fn popup_heuristics() {
...
+        x::ATOM_ATOM,
+        connection.atoms.win_type,
+        &[connection.atoms.win_type_dnd],
+    );
+    connection.set_property::<u32>(
+        discord_dnd,
+        x::ATOM_WM_SIZE_HINTS,
+        x::ATOM_WM_HINTS,
+        &[516, 20, 137, 0, 0, 0, 0, 0, 0],
+    );
```
Supreeeme, line 1750: This can be removed.

Supreeeme (approved):
Thanks! 

That test is failing because windows are only marked as popups if there was a previously focused toplevel to actually parent it to. If you move it to the `popup_heuristics` test (which I'd prefer anyway) it should work.

Supreeeme:
I went ahead and fixed the test myself (https://github.com/Supreeeme/xwayland-satellite/commit/ef24bce5d965360f9d784749874859588b750562). Thanks!

PR #184, opened 2025-06-22
Remove suspicious unwraps added in 799027d

En-En-Code:
Found this regression after crashing xwls about 2 seconds after opening Chaterino7 due to creating a stale surface. The offending `unwrap` is hit since the query `get` call is what finds the stale surface, rather than the construction of the query in `query_one`.

I took a rough look over the rest of that commit, and found a few additional blocks structured the same way and refactored them to also account for the split in query construction and query execution failures.

Doing double `let else` is a smidge unergonomic, but the query has to be held for the duration of the things it gets, so probably not a better way.

### DIFF
--- a/src/server/event.rs
+++ b/src/server/event.rs
@@ -634,45 +634,51 @@ impl Event for client::wl_keyboard::Event {
                 surface,
                 keys,
             } => {
-                if let Some(mut query) = surface.data().copied().and_then(|key| {
+                let Some(mut query) = surface.data().copied().and_then(|key| {
                     state
                         .world
                         .query_one::<(&x::Window, &WlSurface, Option<&OnOutput>)>(key)
                         .ok()
-                }) {
-                    let (window, surface, output) = query.get().unwrap();
-                    state.last_kb_serial = Some((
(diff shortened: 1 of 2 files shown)


src/server/event.rs
```diff
@@ -464,8 +464,11 @@ impl Event for client::wl_pointer::Event {
                     warn!("could not enter surface: stale surface");
                     return;
                 };
+                let Some((surface, role, scale, window)) = query.get() else {
+                    warn!("could not enter surface: stale surface");
+                    return;
+                };
```
Supreeeme, line 470: Idea to avoid double let else:
```rust
let mut query = surface.data().copied().and_then(|e| {
    state
        .world
        .query_one::<(&WlSurface, &SurfaceRole, &SurfaceScaleFactor, &x::Window)>(e)
        .ok()
});
let Some((surface, role, scale, window)) =
    query.as_mut().and_then(|query| query.get())
else {
    warn!("could not enter surface: stale surface");
    return;
};
```
Can also apply elsewhere.

Supreeeme (commented):
Good catch!

PR #198, opened 2025-07-01
Fix `copy_from_wayland` breaking on Nightly

En-En-Code:
Due to [Nightly changes with how `UnixStream` will handle `SIGPIPE`](https://github.com/rust-lang/rust/issues/139956), attempting to [write to a non-socket `UnixStream` will fail](https://github.com/rust-lang/rust/pull/140005#issuecomment-2853932531), a change which appears to be being stabilized.

This resolves the integrations which crash on Nightly by switching to a `PipeWriter`, a fairly new addition to the standard library, which does not have the non-socket restriction.

I think this bumps MSRV for testwl from 1.83.0 to 1.87.0, but testwl doesn't pass `cargo check`, not sure why, not sure how integrations still work, but whatever.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,6 +15,7 @@ all = "deny"
 name = "xwayland-satellite"
 version = "0.6.0"
 edition = "2021"
+rust-version = "1.83.0"
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
--- a/testwl/Cargo.toml
+++ b/testwl/Cargo.toml
@@ -2,9 +2,10 @@
 name = "testwl"
 version = "0.1.0"
 edition = "2021"
+rust-version = "1.87.0"
 
 [dependencies]
-wayland-protocols = { workspace = true, features = ["server", "unstable"] }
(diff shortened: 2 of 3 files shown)


En-En-Code:
Oh, it seems `ubuntu-build` has not been run since the Rust 1.87.0 release. Should probably add a condition in `check_should_build_container`  to keep Rust up-to-date.

Supreeeme:
I didn't realize you could create a pipe with std now. You can switch it to use the std method.
You can also set the `rust-version` field in the `testwl` Cargo.toml. I can work on updating the ci file.

En-En-Code:
Oh, setting `rust-version` is cool but would not resolve the CI.
Yep, `std` has one-way pipes now. I have a refactor branch, but it is already quite a lot. I can take a job of migrating to `std`'s pipe utilities when I finish with it though.

Edit: `rust-version` probably *is* helpful for detcting if the CI should do an update now that I think about it. I set the main project's MSRV as well, so when the refactor to `std`'s pipe happens, it would be an easy confirmation for the CI change.

En-En-Code:
Okay, I have some CI changes to propose, but since GitHub Actions was a problem I used repeated force-pushing to solve, I made a separate branch and PR for it.

PR #208, opened 2025-07-28
Fix `copy_from_wayland` breaking in Nightly + CI tweaks (continues #198)

En-En-Code:
See [the oriiginal explanation](https://github.com/Supreeeme/xwayland-satellite/pull/198#issue-3193403020) for these changes.

Since last time, I discovered that the `testwl` and `macros` would fail to build on their own due to missing feature flags. In `testwl`'s case, the "staging" feature flag was specified by `xwayland-satellite` itself, which is not a major deal since compiling `xwayland-satellite` is necessary for `testwl` to be useful. The `macros` package, on the other hand, needs the "full" feature flag of `syn`, and was only getting it by a *dependency* of `xwayland-satellite`, which had much more potential to break down the line. An additional check in CI for workspace members was added.

The CI now checks for package rebuilds on PRs, along with checking for a modification of the project-wide `rust-version`. Giving each project its own MSRV and rebuilding on a bump only if the combined latest MSRV was bumped felt not worth the additional complexity to me.

Lastly, the script which checks for if a rebuild is needed also determines the MSRV, although nothing it done with it yet. It should not be too hard to set the toolchain to the MSRV in the Dockerfile, which would rebuild the toolchain, and can be done either in this PR or in the future.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,6 +11,7 @@ env:
   REGISTRY: ghcr.io
   IMAGE_NAME: ${{ github.repository }}
   UBUNTU_DOCKERFILE: .github/workflows/ubuntu.dockerfile
+  ROOT_CARGO_TOML: Cargo.toml
 
 permissions:
   contents: read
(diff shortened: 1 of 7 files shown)


En-En-Code:
Oh, I now see why the CI was not previously allowing PRs to update the cache. It works [if you are the owner of the PR](https://github.com/En-En-Code/xwayland-satellite/pull/1), which in hindsight is a reasonable validation. Maybe I can somehow set the `container_path` to point to the PR owner's cache instead.

**WARNING**: I might do more force-pushing to this branch.

En-En-Code:
So there are two solutions I've seen for this issue:
1. Add a Personal Access Token with image push permissions and use the token for Docker authentication.
2. Migrate to `actions/upload-artifact` and `actions/download-artifact` over `docker/login-action` to store the dependency build stage, then use a Personal Access Token to pull an artifact out of a previous workflow run.

I think I prefer the second method as it would allow a PR which bumps MSRV to have its own artifact, with the MSRV in the name, to the artifact being used to test `main`. Either way, I am unable to usefully make more CI progress.

Supreeeme:
Wouldn't you run into artifact storage time limit issues?

En-En-Code:
Oh yeah, you would. So what *would* work then?

I've decided to do some research on Rust CI, since this very much should be a problem other people have already solved and the particular strategy I was attempting here has more and more flaws the longer I stare at it.

Issue #187, opened 2025-06-23
Immediate panic


dbdbc:
### Operating System

Nixos unstable

### Wayland Compositor

Niri unstable 2025-06-19

### Describe the issue.

xwayland-satellite instantly crashes when run. The bug was introduced in commit 799027d1 and still occurs on the latest one 117af56.

Console output of `RUST_BACKTRACE=1 cargo run`:
```
 2025-06-23T23:34:54.445Z INFO  xwayland_satellite > Starting xwayland-satellite version v0.6-10-g117af56
 2025-06-23T23:34:54.491Z INFO  xwayland_process   > xwayland glamor: failed to setup GBM backend, falling back to sw accel
thread 'main' panicked at src/server/mod.rs:992:46:
called `Option::unwrap()` on a `None` value
stack backtrace:
   0: rust_begin_unwind
             at /rustc/e71f9a9a98b0faf423844bf0ba7438f29dc27d58/library/std/src/panicking.rs:665:5
   1: core::panicking::panic_fmt
             at /rustc/e71f9a9a98b0faf423844bf0ba7438f29dc27d58/library/core/src/panicking.rs:76:14
   2: core::panicking::panic
             at /rustc/e71f9a9a98b0faf423844bf0ba7438f29dc27d58/library/core/src/panicking.rs:148:5
   3: core::option::unwrap_failed
             at /rustc/e71f9a9a98b0faf423844bf0ba7438f29dc27d58/library/core/src/option.rs:2009:5
   4: core::option::Option<T>::unwrap
             at /rustc/e71f9a9a98b0faf423844bf0ba7438f29dc27d58/library/core/src/option.rs:972:21
   5: xwayland_satellite::server::ServerState<C>::handle_clientside_events
             at ./src/server/mod.rs:992:21
   6: xwayland_satellite::server::ServerState<C>::run
             at ./src/server/mod.rs:962:9
   7: xwayland_satellite::main
             at ./src/lib.rs:207:9
   8: xwayland_satellite::main
             at ./src/main.rs:8:5
   9: core::ops::function::FnOnce::call_once
             at /rustc/e71f9a9a98b0faf423844bf0ba7438f29dc27d58/library/core/src/ops/function.rs:250:5
```

### Steps to reproduce

run xwayland-satellite in debug or release mode

### Please confirm you have reproduced this on the latest commit.

- [x] I have reproduced this on the latest commit

senekor:
[...]

PR #214, opened 2025-08-03
Use typestate to eliminate `None` connections

En-En-Code:
I have been thinking about #187 , since when I looked into it, I diagnosed the root problem as `server_state.run()` assuming `xstate` was not `None`, even though at the beginning of the program it necessarily had to be while managing XWayland's initialization. This felt to me like a design flaw, so after a little tinkering, I found a way to type the pre-initialized and post-initialized states of `ServerState`

The first commit is a refactor which seperates `connection` from the rest of `ServerState`'s contents so the rest of the server state can be transferred when the connection type is changed. Notably, this relies on the `Dispatch` impls not using `connection`, but Wayland events should never be touching the X connection anwyay. This is not a complex change but it does nothing on its own.

The second commit defines `EarlyConnection`, the type which serves as the minimal filler connection before the true connection is initialized and substituted in with `upgrade_connection`. Since this involves type changes, a big chunk of `main` also needed refactors, but it also removed a lot of the conditional checks jammed into the original `main` loop.

The third commit is the payoff. `ServerState::connection` and `InnerServerState::client` can now never be `None`, and so all of the `Option`s and `unwrap` associated with them can be removed. I also wanted [a conditional debug statement](https://github.com/Supreeeme/xwayland-satellite/commit/9872aa9519ecb05c7173fd1ba7f46f668c276c98#diff-1b33b1af1a8e35c4f324cac560774fc81865de0181472f33fd544f199e8cf931R153), so I ended up also added a return type to `set_window_dims`, which required additional changes.

The last commit gets the integration tests to work again, which in commits 2 & 3 always abort (with the `IO Safety violation: owned file descriptor already closed` message) due to [the `xcb::Connection::connect_to_fd*` functions not being sound](https://github.com/rust-x-bindings/rust-xcb/issues/167) (at least to me it is unsound. I have a minimal example and will make the issue and RustSec PR). The fix should prevent that abort from also spuriously occuring in integration runs.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -54,7 +54,7 @@ pub fn main(mut data: impl RunData) -> Option<()> {
     info!("Starting xwayland-satellite version {version}");
 
     let socket = ListeningSocket::bind_auto("xwls", 1..=128).unwrap();
-    let mut display = Display::<RealServerState>::new().unwrap();
+    let mut display = Display::new().unwrap();
     let dh = display.handle();
     data.created_server();
 
@@ -203,7 +203,7 @@ pub fn main(mut data: impl RunData) -> Option<()> {
             xstate.handle_events(&mut server_state);
         }
 
-        display.dispatch_clients(&mut server_state).unwrap();
+        display.dispatch_clients(server_state.inner_mut()).unwrap();
         server_state.run();
         display.flush_clients().unwrap();
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -16,7 +16,7 @@ use xcb::x;
 pub trait XConnection: Sized + 'static {
     type X11Selection: X11Selection;
 
-    fn set_window_dims(&mut self, window: x::Window, dims: PendingSurfaceState);
+    fn set_window_dims(&mut self, window: x::Window, dims: PendingSurfaceState) -> bool;
     fn set_fullscreen(&mut self, window: x::Window, fullscreen: bool);
     fn focus_window(&mut self, window: x::Window, output_name: Option<String>);
     fn close_window(&mut self, window: x::Window);
@@ -134,8 +134,7 @@ pub fn main(mut data: impl RunData) -> Option<()> {
         }
     };
 
-    let mut server_state = EarlyServerState::new(dh, data.server());
-    server_state.connect(connection);
+    let mut server_state = EarlyServerState::new(dh, data.server(), connection);
     server_state.run();
 
     // Remove the lifetimes on our fds to avoid borrowing issues, since we know they will exist for
@@ -177,8 +176,7 @@ pub fn main(mut data: impl RunData) -> Option<()> {
         display.flush_clients().unwrap();
     }
 
-    let mut xstate: Option<XState> = None;
-    let xstate = xstate.insert(XState::new(xsock_wl.as_fd()));
+    let mut xstate = XState::new(xsock_wl.as_fd());
     let mut reader = BufReader::new(&ready_rx);
     {
         let mut display = String::new();
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -59,6 +59,11 @@ pub fn main(mut data: impl RunData) -> Option<()> {
     data.created_server();
 
     let (xsock_wl, xsock_xwl) = UnixStream::pair().unwrap();
+    // XCB takes responsibility for cleaning up this FD, but since connecting takes a RawFd at the
+    // FFI level, see (`XState::new`), `xsock_wl`'s destructor also closes the FD, leading the FD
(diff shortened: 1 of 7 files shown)


src/server/mod.rs
```diff
@@ -393,6 +393,33 @@ struct GlobalOutputOffset {
     y: GlobalOutputOffsetDimension,
 }
 
+/// The state of the X11 connection before XState has been fully initialized.
+/// It implements XConnection minimally, gracefully doing nothing but logging the called functions.
+pub struct EarlyConnection<S: X11Selection + 'static> {
```
Supreeeme, line 396: I think it's clearer if this is just `NoConnection`

src/lib.rs
```diff
@@ -154,67 +151,82 @@ pub fn main(mut data: impl RunData) -> Option<()> {
...
-            }
+        display.dispatch_clients(server_state.inner_mut()).unwrap();
+        server_state.run();
+        display.flush_clients().unwrap();
+    }
 
-            #[cfg(not(feature = "systemd"))]
-            info!("Systemd support disabled.");
-        }
+    let mut xstate: Option<XState> = None;
```
Supreeeme: Why is this an Option?

src/lib.rs
```diff
@@ -59,6 +59,11 @@ pub fn main(mut data: impl RunData) -> Option<()> {
     data.created_server();
 
     let (xsock_wl, xsock_xwl) = UnixStream::pair().unwrap();
+    // XCB takes responsibility for cleaning up this FD, but since connecting takes a RawFd at the
+    // FFI level, see (`XState::new`), `xsock_wl`'s destructor also closes the FD, leading the FD
+    // being closed twice. This mainly caused problems in the integration tests, where `xsock_wl`'s
+    // destructor would be run after the descriptor was freed, leading to an opaque abort message.
```
Supreeeme, line 65: Could add a link to the xcb issue here.

src/server/mod.rs
```diff
@@ -510,31 +513,289 @@ impl<C: XConnection> ServerState<C> {
...
+        self.inner.set_popup(window, is_popup)
+    }
+
+    pub fn set_win_title(&mut self, window: x::Window, name: WmName) {
+        self.inner.set_win_title(window, name);
+    }
+
+    pub fn set_win_class(&mut self, window: x::Window, class: String) {
+        self.inner.set_win_class(window, class)
+    }
```
Supreeeme, line 601: I don't love these, just another layer of indirection. I don't know if I feel strongly enough about it to need these to be inlined here. I can be swayed.

An alternate idea would to be to have ServerState Deref to InnerServerState.

src/server/mod.rs
```diff
@@ -510,31 +513,289 @@ impl<C: XConnection> ServerState<C> {
...
+        self.inner.handle_activations()
+    }
+
+    fn calc_global_output_offset(&mut self) {
+        self.inner.calc_global_output_offset()
+    }
+
+    // create_role_window is only called in a Dispatch impl, a trait impl necessarily moved to be
+    // on the ineer value. create_toplevel and create_popup are only called by that create_role_window
+    // call, so none of them need wrapper functions.
```
Supreeeme, line 800:
```suggestion
```

Supreeeme (commented):
This looks pretty good, just a few comments. I thought about doing something like this at some point but determined I didn't care enough to do all the work for it lol, so thanks!

src/server/mod.rs
En-En-Code: I took the option of implementing `Deref` and `DerefMut`. It really does clean up so much of this PR. All of the boilerplate functions gone, and a lot of noisy `inner` accesses in other places to be implied and removed (not all, of course. When you need the `InnerServerState` mutable and the connection at the same time, then you still have to destructure the state).
I think I am also going to refactor the unit test fixture to also have an inner type which can be accessed by `Deref`. I should be able to assert every method of `XConnection` can be invoked on `NoConnection` and still run afterwards.

src/server/event.rs
```diff
@@ -624,6 +627,7 @@ impl Event for client::wl_pointer::Event {
 
 impl Event for client::wl_keyboard::Event {
     fn handle<C: XConnection>(self, target: Entity, state: &mut ServerState<C>) {
+        let state = state.deref_mut();
```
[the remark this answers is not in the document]
En-En-Code: Nice catch. Many are them were. The one in particular you selected is not one of them, since different fields of  `InnerServerState` need to be borrowed at the same time, with at least one of them being a mutable borrow. Being reassigning `state` to the inner type, it allows the compiler to perform field destructuring, which it cannot perform on the outer `ServerState`.

Supreeeme:
Thanks!

PR #222, opened 2025-08-21
Rewrite CI from ground up

En-En-Code:
Continues on #198 and #208. Third time's the charm, surely. ;)

Decided to tear out the Docker container stuff, as there were multiple nightmarish problems with describing to GitHub Actions that sometimes it would need to pull down the container and rebuild it other times and with needing to extract the container into a form which could be understood by Docker and with copying the actual repo contents into the container efficiently. I had something that I thought was good progress, but despite cutting an entire redundant build of all of the Rust dependencies (because `build` and `test` were built with different profiles), it still took as long as the original CI.

Enter the rewrite. Yes, it downloads all of the dependencies everytime, (it can take 20 seconds, it can take 2 minutes, 20 seconds), but I cannot complain about keeping them up to date. Looks like a pretty standard Rust CI, testing the MSRV toolchain and even build dependency caching (I do not know if this project has enough activity for that cache to actually be useful, but I figure it cannot hurt anything).


En-En-Code:
> I am not sure how I feel about getting rid of the dockerfile and installing the deps every CI run. If we're using a github action to set up the rust toolchain then I don't see why the dockerfile needs to go.

I think I made that change to assert the deps are up-to-date on any given run, but pulling down a smaller image (since the Rust toolchain would not be installed) could be a fine strategy. I wonder if you could also then add a special job that updates them, say, once before a new release.

En-En-Code:
Oh my goodness. A freaking check mark. The Dockerfile is back, for dependency caching. It rebuilds only on pushes to main which change the Dockerfile or annotated tags on HEAD for dependency updating. This excludes PRs which change the Dockerfile, but I believe needing further Dockerfile changes would be rare at this point.
Also got rid of the cache step since it needed `node` on the container and I was not going to figure out what else.

PR #223, opened 2025-08-22
Bump `xcb` to 1.6.0

En-En-Code:
Now that `xcb` has deprecated `Connection::connect_to_fd_with_extensions` for the I/O safety problems it caused and added `Connection::connect_with_fd_and_extensions` to serve as the sound alternative, we bump to 1.6.0 and switch to it. Nicely get to remove the hacky line we were previously relying on for correct behavior.

Also, I have several small commits I want to contribute after the CI stuff gets resolved. Should I split them into seperate PRs or can I put them all in one?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1153,9 +1153,9 @@ dependencies = [
 
 [[package]]
 name = "xcb"
-version = "1.5.0"
+version = "1.6.0"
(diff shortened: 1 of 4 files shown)


Supreeeme:
> Should I split them into seperate PRs or can I put them all in one?

If they are unrelated, it will be easier for me to review if they are in separate PRs.

Issue #91, opened 2025-01-17
Clipboard stops syncing between wayland and xwayland-satellite after running for a long time


ToxicMushroom:
I do have a log but its like 3.2Gb created with `WAYLAND_DEBUG=1 RUST_LOG=debug xwayland-satellite 2> xwayland-satellite.log`

Both wayland -> xwayland-satellite and xwayland-satellite -> wayland just stopped syncing.

maybe i can get a smaller log later.

If there is something I could search for in the logfile to help out let me know :)

ToxicMushroom:
`tail -f xwayland-satellite.log | grep data_source` doesn't log any new data sources when copying something inside an xwayland app (tried vesktop and xclip).

Though I do have data source logs from 2-3 hours ago (was still working around that time iirc)

Supreeeme:
Running with `RUST_LOG=xwayland_satellite::xstate::selection` will give you a log with just the interesting stuff.

zeroeightysix:
I think this has been fixed since the clipboard refactor?

crescentrose:
I have a similar issue, except clipboard sync from X11 to Wayland just does not work at all for me. 

When copying from an X11 window, I get this:

```
xwayland-satellite[162788]:  2025-03-22T13:07:45.467Z WARN  xwayland_satellite::xstate        > unrecognized message: Atom { res_id: 253 }
xwayland-satellite[162788]:  2025-03-22T13:08:03.706Z DEBUG xwayland_satellite::xstate::selection > new selection owner: Window { res_id: 18874368 }
xwayland-satellite[162788]:  2025-03-22T13:08:03.706Z DEBUG xwayland_satellite::xstate::selection > selection notify requestor: Window { res_id: 2097157 } target: TARGETS
xwayland-satellite[162788]:  2025-03-22T13:08:03.706Z DEBUG xwayland_satellite::xstate::selection > got targets: ["TARGETS", "SAVE_TARGETS", "UTF8_STRING", "text/plain;charset=utf-8", "text/plain;charset=UTF-8"]
xwayland-satellite[162788]:  2025-03-22T13:08:03.706Z DEBUG xwayland_satellite::xstate::selection > Clipboard set from X11
```

So seemingly, the copy operation is successful, but then if I try pasting, even just with `wl-paste`, I get this message:

```
xwayland-satellite[162788]:  2025-03-22T13:08:05.238Z WARN  xwayland_satellite::xstate::selection > selection notify fail?
```

[...]

PR #243, opened 2025-09-28
transfer via INCR Wayland selections larger than the X server's `maximum-request-length`

En-En-Code:


Supreeeme:
I wonder if this fixes #91.

src/xstate/selection.rs
```diff
@@ -162,12 +162,59 @@ impl Selection {
...
+        let incr_data = self.incr_data.as_mut()?;
+        let range_start = incr_data.range.start;
+        let range_len = std::cmp::min(incr_data.max_req_bytes, incr_data.range.end - range_start);
+
+        if let Err(e) = connection.send_and_check_request(&x::ChangeProperty {
+            mode: x::PropMode::Append,
+            window: incr_data.target_window,
+            property: incr_data.property,
+            r#type: incr_data.target_type,
+            data: &incr_data.data[range_start..][..range_len],
```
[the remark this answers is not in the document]
En-En-Code: Yeah, this maybe was me getting a little too cute. Slicing slices is the kind of thing you might see in Python code a lot (but I don't write Python, so I still don't know why I'd think to do it this way).

En-En-Code:
I do not believe this is related to #91. It only changes Wayland -> X so it would not be a full solution anyways. About the worst I have been able to consistently produce is a few tenths of a second of lag from X -> Wayland and only from some X apps.

src/xstate/selection.rs
```diff
@@ -162,12 +162,61 @@ impl Selection {
...
+            );
+            self.incr_data = None;
+        } else {
+            debug!(
+                "received some incr data for {}",
+                get_atom_name(connection, incr_data.target_type)
+            );
+            incr_data.start = incr_end;
+        }
+        Some(true)
```
Supreeeme, line 213: This function appears to return `Some(true)` in all cases, so the return value seems a bit pointless.

Supreeeme (approved):

src/xstate/selection.rs
En-En-Code: You're missing the `?` operator in the first line that causes a return value of `None` to be possible.

Supreeeme: That I did. Does it need to be an Option though? Seems like it could just be bool.

Supreeeme: Hmm, I'm not sure how the return values wouldn't convey they same information? Wouldn't they both be true if there was incr data to process and false otherwise?

En-En-Code:
Tacked on the additonal fix here since if it were on its own branch, the changes would conflict with those on this one. I like not needing to deal with conflicts when I can see them coming ahead of time.

PR #250, opened 2025-10-03
bump rustix to version 1

En-En-Code:


src/server/tests.rs
```diff
@@ -1392,11 +1396,15 @@ fn copy_from_wayland<T: SelectionTest>() {
             s.spawn(|| {
                 let pollfd = unsafe { BorrowedFd::borrow_raw(f.testwl.poll_fd().as_raw_fd()) };
                 let mut pollfd = [PollFd::from_borrowed_fd(pollfd, PollFlags::IN)];
-                if poll(&mut pollfd, 100).unwrap() == 0 {
+                let timeout = Some(&Timespec {
+                    tv_sec: 0,
+                    tv_nsec: 100_000_000,
```
Supreeeme, line 1401:
```suggestion
                    tv_nsec: std::time::Duration::from_millis(100).as_nanos()
```
This could also go in a `const`.

Cargo.toml
```diff
@@ -2,11 +2,14 @@
...
 [workspace.dependencies]
-wayland-client = "0.31.2"
-wayland-protocols = "0.32.0"
-wayland-scanner = "0.31.1"
-wayland-server = "0.31.1"
-rustix = "0.38.31"
+wayland-client = "0.31.11"
+# We don't directly use wayland-cursor, but we need to set its version so
+# smithay-client-toolkit does not depend on an older version with rustix < 1
+wayland-cursor = "0.31.11"
```
Supreeeme, line 8: I don't see why this would be the case? smithay-client-toolkit 0.20 depends on rustix 1.0.7.

En-En-Code: It was from the older `wayland-cursor` dependency (from before `rustix` 1) still lingering in the Cargo.lock and this was the easiest way I found to massage the Cargo.lock into bumping `wayland_cursor` without bumping unrelated things. It indeed does not need to be in the Cargo.toml anymore.

PR #259, opened 2025-10-26
Allow enabling extensions on the X server

Ednaordinary:
xwayland-satellite does not allow specifically enabling extensions. This PR adds support for it to do so. This is particularly helpful for OpenGL with the GLX extension. I've found this helpful when using xwayland-satellite in combination with waypipe, where GLX seems to not be enabled by default

### DIFF
--- a/tests/integration.rs
+++ b/tests/integration.rs
@@ -86,6 +86,10 @@ impl xwls::RunData for TestData {
     fn max_req_len_bytes(&self) -> Option<usize> {
         Some(500)
     }
+    fn ext_add(&self) -> Vec<&str> {
+        let ext: Vec<&str> = Vec::new();
+        ext
+    }
 }
 
 struct Fixture {


Ednaordinary:
It may also be beneficial to allow enabling -eglstream, but the server I'm testing on has broken Nvidia drivers right now. The server can also use -extension instead of +extension to disable extensions, but I'm not sure if that really needs to be implemented. I can add that pretty easily if it would be beneficial

Supreeeme:
I'm a little surprised you even need this. GLX seems to work for me without enabling this, and adding `+extension GLX` doesn't actually seem to actually change anything, even over waypipe. Does GLX work with Xwayland locally for you?

Ednaordinary:
Ah, apologies, I seem to have tricked myself into thinking this was needed. I was running blender via nix shell which for whatever reason gives me `GLX_ARB_create_context not available`, then thought adding the GLX extension would fix this, it did not but using the regular blender binary works. At some point that switched to me thinking `+extension GLX` helped in some way. It may still be helpful to allow extension loading for those which don't load by default, but I'll leave it up to you whether or not to merge

(My environment is acting differently than I remember it before, though that might be something unrelated)

Ednaordinary:
Nevermind, I tried testing on a different machine and found it does indeed make a difference in some cases.

In a local terminal:
```
waypipe --socket /tmp/meow.sock -c zstd=10 client
```
In another local terminal:
```
ssh -R /tmp/meow.sock:/tmp/meow.sock user@a-different-machine
waypipe --socket /tmp/meow.sock -c zstd=10 --no-gpu --threads 16 server
xwayland-satellite :2
```
Then on a terminal on user@a-different-machine:
```
export DISPLAY=:2
glxgears
```
Which, depending on the set up of the computer, can error with `Error: couldn't get an RGB, Double-buffered visual`. Killing `xwayland-satellite` and restarting with `xwayland-satellite :2 +extension GLX` allows running. I'm unsure exactly why, though if I had to guess it has something to do with GLX shared object libraries not being loaded (?)
image

Note that without specifying `__GLX_VENDOR_LIBRARY_NAME=nvidia`, this will fall back to llvmpipe based mesa software rendering

Supreeeme:
You pass `--no-gpu` to that command, which from the waypipe help:
```
 -n, --no-gpu                Block protocols using GPU memory transfers (via DMABUFs)
```
Which kinda makes sense that blocking dmabufs would disable GLX because I imagine that's what Xwayland uses for GLX. Now the question is, is force enabling the extension something we want? I can't think of an immediate reason why not, so maybe.

Supreeeme:
Sorry for just sitting on this, but I'm probably going to go with the approach in #342.

Issue #252, opened 2025-10-16
xwayland-satellite panics when opening beta Steam with clipboard not empty


LuckShiba:
### Operating System

Fedora 42

### Wayland Compositor

Niri

### Describe the issue.

Whenever I open the Steam beta client (both system and Flatpak package), xwayland-satellite is panicking if my clipboard is not empty due to an unwrap in an Err value. This is not happening in the stable Steam client yet.

```
thread 'main' panicked at src/xstate/selection.rs:388:22:
called `Result::unwrap()` on an `Err` value: X(Window(ValueError { response_type: 0, error_code: 3, sequence: 214, bad_value: 39845933, minor_opcode: 0, major_opcode: 18, pad: 1 }), Some("x::ChangeProperty"))
```

Relevant logs:
[xwayland-satellite.log](https://github.com/user-attachments/files/22955909/xwayland-satellite.log)

### Steps to reproduce

1. Have something in your clipboard
2. Open Steam beta client
3. xwayland-satellite will panic
4. Run `wl-copy --clear`
5. xwayland-satellite won't panic when launching Steam with a clear clipboard

### Please confirm you have reproduced this on the latest commit.

- [x] I have reproduced this on the latest commit

En-En-Code:
Wait, why is there just not proper error handling here? Should be easy to resolve at least. Looking around I see a few other `unwraps` that should go as well if I'm cleaning this up.

LuckShiba:
> Wait, why is there just not proper error handling here? Should be easy to resolve at least. Looking around I see a few other `unwraps` that should go as well if I'm cleaning this up.

I tried your branch, now it's another unwrap, on the line 794

```
Failed to set targets for selection request: X(Window(ValueError { response_type: 0, error_code: 3, sequence: 213, bad_value: 20971565, minor_opcode: 0, major_opcode: 18, pad: 1 }), Some("x::ChangeProperty"))

thread 'main' panicked at src/xstate/selection.rs:794:26:
called `Result::unwrap()` on an `Err` value: X(Window(ValueError { response_type: 0, error_code: 3, sequence: 215, bad_value: 20971565, minor_opcode: 0, major_opcode: 25, pad: 1 }), Some("x::SendEvent"))
```

LuckShiba:
update: with your patches and with https://github.com/LuckShiba/xwayland-satellite/commit/87d3e84579e37a2159f2dee7b1a98f9f972e1048 the problem seems to be solved.

[...]

PR #260, opened 2025-10-31
Add #252 fix + refactors

En-En-Code:
I pulled down LuckShiba's patch (since I was working on other clipboard fixes), but noticed that it was not formatted with `rustfmt`. Since I would have to make a commit to fix that, I also used the commit to do some additonal refactoring, at which point I realized the purpose of the two commits diverged heavily enough from the rest of the branch to be its own PR.

### DIFF
--- a/src/xstate/mod.rs
+++ b/src/xstate/mod.rs
@@ -590,7 +590,10 @@ impl XState {
                     }
                 }
             }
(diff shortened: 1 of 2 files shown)


Supreeeme:
The 2nd commit is doing a few unrelated things, can you split it up?

PR #262, opened 2025-11-01
Fix 2 issues with Git Gui menus

### LINKED ISSUE #253: git gui menus do not work properly
### Operating System

Arch Linux

### Wayland Compositor

Niri

### Describe the issue.

When launching `git gui` the tk gui wrapper for git `/usr/lib/git-core/git-gui` the menu system is not working 

### Steps to reproduce

run git gui, click anything in the menus and instead of showing up in the right spot they open in dedicated windows.

En-En-Code:
Partially resolves #253. Apparently there is also a scaling issue, but I do not understand it and the issue author has not responded in some time.

### DIFF
--- a/src/server/event.rs
+++ b/src/server/event.rs
@@ -460,39 +460,45 @@ impl Event for client::wl_pointer::Event {
         // Niri). Other compositors do not run into this problem because they appear to not send
         // wl_pointer.enter until the user actually moves the mouse in the popup.
 
-        let handle_pending_enter = |target: Entity, state: &mut ServerState<C>, event_str: &str| loop {
-            let Ok(pe) = state.world.get::<&PendingEnter>(target) else {
-                return true;
-            };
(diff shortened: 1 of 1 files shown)


src/server/event.rs
```diff
@@ -460,6 +460,40 @@ impl Event for client::wl_pointer::Event {
...
+            };
+            let PendingEnter(client::wl_pointer::Event::Enter {
+                serial,
+                surface,
+                surface_x,
+                surface_y,
+            }) = pe.deref()
+            else {
+                unreachable!();
+            };
```
Supreeeme: Is this else needed? This seems like an irrefutable pattern.

En-En-Code: `PendingEnter` is a wrapper around the `wl_pointer::Event` enum, which has more than just the `Enter` variant, even if that is all we store in it. Just I could redefine `PendingEnter` if we really don't it.

PR #281, opened 2025-11-19
fix: popup position

busyoGG:
Maybe this can fix some popup menu position issues, at least for WeChat.

I found WeChat's menu position is set after it's mapped.

The position starts at (0,0). When win.mapped is set to true, the reconfiguration path is bypassed, though the dims now contain the genuine coordinates.

``` rust
if !win.mapped {
    win.attrs.dims = dims;
    return;
}
```

screenshot-2025-11-19_14-15-18

There are still other popup detection issues, but those are related to the heuristic function and are outside the scope of this PR. Therefore, this PR can only partially address the related problems."

### DIFF
--- a/src/server/mod.rs
+++ b/src/server/mod.rs
@@ -941,13 +941,14 @@ impl<S: X11Selection + 'static> InnerServerState<S> {
         };
         if dims == win.attrs.dims {
             return;
-        } else {
+        } else if win.attrs.is_popup {
             win.attrs.dims = dims;
         }
 
         debug!("Reconfiguring {:?} {:?}", event.window(), dims);
 
         if !win.mapped {
+            win.attrs.dims = dims;
             return;
         }
 


En-En-Code:
Thanks for taking the time to sift the logs, find the cause, and present a reasonable solution!

This change does not consider the case of top-levels (windows where `win.attrs.is_popup == false`), which, unlike pop-ups, should not be moved after being mapped (but still can before being mapped). It might just be a technicality, since the `ConfigureNotifyEvent` for top-levels should not send data where `event.x() != win.attrs.dim.x` (and likewise for `y`), but that is an additional assumption this code would be relying on. This is just my two cents. Ultimately, you would have to wait for @Supreeeme's opinion on this change.

src/server/mod.rs
```diff
@@ -941,12 +941,10 @@ impl<S: X11Selection + 'static> InnerServerState<S> {
...
         if dims == win.attrs.dims {
             return;
-        }
-        debug!("Reconfiguring {:?} {:?}", event.window(), dims);
-        if !win.mapped {
+        } else {
             win.attrs.dims = dims;
-            return;
         }
+        debug!("Reconfiguring {:?} {:?}", event.window(), dims);
```
Supreeeme: I don't think the `win.mapped` check should go away here. Honestly you can just keep the if statement and move `win.attrs.dims = dims` right before it.

Supreeeme (commented):
I agree with En-En in that we need to take into account that toplevel positions should be ignored, but I otherwise like this change, seems like an unfortunate oversight I missed. This could also use a new test if you're up for it, but I won't require you to write one.

src/server/mod.rs
busyoGG: OK, I'll re-add `win.mapped` check.

busyoGG:
> I agree with En-En in that we need to take into account that toplevel positions should be ignored, but I otherwise like this change, seems like an unfortunate oversight I missed. This could also use a new test if you're up for it, but I won't require you to write one.

I'm not sure how to write the appropriate test code. I've made some attempts, but the results have been unsatisfactory.

En-En-Code:
The case we want the test, where the reconfigure event is sent between the window being mapped and having a `SurfaceRole`, would require modifying the `create_popup` test helper to allow reconfigure events to occur within that timespan. This definitely requires some amount of thoughtful design, likely to `PopupBuilder`.

En-En-Code:
Alright, I have written the [new unit tests](https://github.com/En-En-Code/xwayland-satellite/compare/pr-281-tests) to test your changes against. Yeah, it was pretty involved. I would not expect someone without little experience in the codebase to figure this out in a reasonable amount of time.

src/server/mod.rs
```diff
@@ -941,10 +941,13 @@ impl<S: X11Selection + 'static> InnerServerState<S> {
         };
         if dims == win.attrs.dims {
             return;
+        } else {
+            win.attrs.dims = dims;
```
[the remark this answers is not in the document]
busyoGG: Modified. Keep the original logic for toplevel and apply the new logic only for popup.

PR #283, opened 2025-11-22
fix: problem unwrap in `wl_pointer::Event` (again)

En-En-Code:
I triggered the unwrap in the `Motion` case while using the color picker in Krita. If you hold down the mouse such that the color comparison pop-up opens and flick to it, you can sometimes enter and motion on a stale surface. Somewhat inconsistent to trigger.

I also removed the unwrap in the `Button` event since it is logically equivalent to the other unwrap. A few logging statements also got changed to give wl_surface details.


Supreeeme:
Looks good, other than `cargo fmt` :eyes:

Issue #294, opened 2025-12-02
Regression: ardour8 + yabridge + VST plugin spawns as popup instead of normal window


YaLTeR:
### Operating System

Fedora 43

### Wayland Compositor

niri

### Describe the issue.

I use ardour8 + yabridge + wine with "emulate virtual desktop" (that doesn't change the outcome here).

After e719ce746a0f36c33d1449bd6cf1d8b164bd1712, the VST window now gets embedded into Ardour (or wine's virtual desktop) instead of showing up as a separate window:

<img width="1369" height="896" alt="Image" src="https://github.com/user-attachments/assets/19859abb-3528-45fc-8a8a-164d26639a82" />

```
WM_HINTS(WM_HINTS):
		Client accepts input or input focus: True
		Initial state is Normal State.
		bitmap id # to use for icon: 0x16002e2
		bitmap id # of mask for icon: 0x16002e3
		window id # of group leader: 0x1600001


WM_WINDOW_ROLE(STRING) = "plugin_ui"
_NET_WM_SYNC_REQUEST_COUNTER(CARDINAL) = 23070398
_NET_WM_WINDOW_TYPE(ATOM) = _NET_WM_WINDOW_TYPE_UTILITY
_NET_WM_USER_TIME(CARDINAL) = 32540685
_NET_WM_USER_TIME_WINDOW(WINDOW): window id # 0x16006bd
WM_CLIENT_LEADER(WINDOW): window id # 0x1600001
_NET_WM_PID(CARDINAL) = 131469
WM_LOCALE_NAME(STRING) = "en_US.UTF-8"
WM_CLIENT_MACHINE(STRING) = "toolbx"
WM_NORMAL_HINTS(WM_SIZE_HINTS):
		program specified location: 0, 0
		program specified minimum size: 1190 by 769
		program specified maximum size: 1190 by 769
		window gravity: NorthWest
WM_PROTOCOLS(ATOM): protocols  WM_DELETE_WINDOW, WM_TAKE_FOCUS, _NET_WM_PING, _NET_WM_SYNC_REQUEST
WM_CLASS(STRING) = "ardour-8.12.0", "Ardour-8.12.0"
WM_ICON_NAME(STRING) = "Superior Drummer 3: Superior Drummer 3 (by Toontrack) [VST3]"
_NET_WM_ICON_NAME(UTF8_STRING) = "Superior Drummer 3: Superior Drummer 3 (by Toontrack) [VST3]"
[...]

Issue #293, opened 2025-12-02
REAPER + Yabridge + VST menus dont show as popups


GoranKovac:
### Operating System

Arch

### Wayland Compositor

Hyprland

### Describe the issue.

Similar to https://github.com/Supreeeme/xwayland-satellite/issues/221

When opening context menus on VST plugins (with yabridge setup) they spawn at center of the screen and not as popups. Here is an example with Superior Drummer 3 but many plugins suffer from the same issue:

https://github.com/user-attachments/assets/d05e7f4c-56d6-4e42-b73b-1fb4f52507ea

Log from `RUST_LOG=debug` (starting reaper, inserting superior drummer, right click on element)
[log.txt](https://github.com/user-attachments/files/23877566/log.txt)

xprop:
```
WM_STATE(WM_STATE):
		window state: Normal
		icon window: 0x0
_NET_WM_ICON(CARDINAL) = 	Icon (32 x 32):
	▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄
	▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄
	▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄
	▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄
	▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄
	▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄
	▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄
	▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄
	▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄
[...]

Issue #291, opened 2025-12-01
No mouse cursor in some wine screens (exe installers)


GoranKovac:
### Operating System

Arch

### Wayland Compositor

Hyprland

### Describe the issue.

When trying to install apps via wine there is no mouse cursor and you cant click or do anything to proceed with installation.

Only thing that pops up in satellite output it:
```
 2025-12-01T19:19:09.530Z WARN  xwayland_satellite::server > Window Window { res_id: 16777222 } was not an active toplevel, not setting as parent
```
This issue does not happen in hyprlands implementation of xwayland

https://github.com/user-attachments/assets/3d0254b2-e8e7-413e-91ea-438bd1675b77

https://www.tseaudio.com/software/tseBOD

```
WM_STATE(WM_STATE):
		window state: Normal
		icon window: 0x0
_NET_WM_ICON(CARDINAL) = 	Icon (32 x 32):	

_NET_WM_STATE(ATOM) = _NET_WM_STATE_SKIP_TASKBAR
_NET_WM_NAME(UTF8_STRING) = "Setup - TSE BOD"
WM_ICON_NAME(STRING) = "Setup - TSE BOD"
WM_NAME(STRING) = "Setup - TSE BOD"
WM_HINTS(WM_HINTS):
		Client accepts input or input focus: False
		Initial state is Normal State.
		bitmap id # to use for icon: 0x1000126
		bitmap id # of mask for icon: 0x1000128
		window id # of group leader: 0x1200006
_NET_WM_WINDOW_TYPE(ATOM) = _NET_WM_WINDOW_TYPE_NORMAL
WM_TRANSIENT_FOR(WINDOW): window id # 0x1200006
_MOTIF_WM_HINTS(_MOTIF_WM_HINTS) = 0x3, 0x2c, 0x3a, 0x0, 0x0
WM_NORMAL_HINTS(WM_SIZE_HINTS):
		program specified location: 2951, 553
		program specified minimum size: 497 by 360
		program specified maximum size: 497 by 360
		window gravity: Static
_NET_WM_USER_TIME_WINDOW(WINDOW): window id # 0x1000039
XdndAware(ATOM) = BITMAP
_NET_WM_PID(CARDINAL) = 141255
WM_LOCALE_NAME(STRING) = "en_US.UTF-8"
WM_CLIENT_MACHINE(STRING) = "archlinux"
[...]

Issue #280, opened 2025-11-17
Popup follow behavior for Wine X11 apps (e.g., CLIP STUDIO PAINT)


Unk-now:
This issue is a specific case related to the general popup problem discussed in #221.

Dear developers,

Thank you for your ongoing work on xwayland-satellite.

I investigated the menu behavior of certain X11 applications running through Wine (e.g., CLIP STUDIO PAINT) and would like to share my findings.

---

### 1. Issue Overview
- When opening menus in Wine-based applications, submenus (popups) do not appear next to the parent menu item, but instead show up as independent windows in the center of the screen.

### 2. Expected Behavior
- When the mouse hovers over a menu item, the corresponding submenu should appear directly below the item and follow the cursor.

### 3. Analysis
- Using `xprop` on the problematic popups, I observed the following common properties:
  - `WM_TRANSIENT_FOR` is set.
  - `_NET_WM_NAME` and `WM_NAME` are empty.

> This suggests that windows with these characteristics could be reliably treated as popups.

#### Example xprop output
```text
_MOTIF_WM_HINTS(_MOTIF_WM_HINTS) = 0x3, 0x24, 0x0, 0x0, 0x0
_NET_WM_NAME(UTF8_STRING) = 
WM_ICON_NAME(STRING) = 
WM_NAME(STRING) = 
WM_HINTS(WM_HINTS):
    Client accepts input or input focus: False
    Initial state is Normal State.
    bitmap id # to use for icon: 0xc03fee
    bitmap id # of mask for icon: 0xc03ff0
    window id # of group leader: 0xe0000a
_NET_WM_WINDOW_TYPE(ATOM) = _NET_WM_WINDOW_TYPE_DIALOG
WM_TRANSIENT_FOR(WINDOW): window id # 0xe0000a
...
```

### 4. Attempted Fix
- Added the following heuristic in `src/server/mod.rs` within `create_role_window`:

```rust
let window_data = data.get::<&WindowData>().unwrap();
if window_data.attrs.is_popup {
    popup_for = self.last_hovered.or(self.last_focused_toplevel);
} else {
    let attrs = &window_data.attrs;

    let title_is_effectively_empty = match &attrs.title {
        None => true,
        Some(title_enum) => match title_enum {
            WmName::WmName(contents) => contents.is_empty(),
            WmName::NetWmName(contents) => contents.is_empty(),
        },
    };

    if attrs.transient_for.is_some() && title_is_effectively_empty {
[...]

PR #298, opened 2025-12-02
Several Popup Heuristics Fixes

En-En-Code:


Supreeeme:
> When `guess_is_popup` was first written, the `skip_taskbar` check was
put after the loop over `window_types`. Because of how `window_types`
was defined, however, this codepath would only ever be used if
`window_types` was not empty and none of its values were checked for
window atoms, which is quite rare.

For the record, I think I might've done this intentionally - at least, I feel like `SKIP_TASKBAR` should really only be used as a last resort if we can't figure out what kind of window it should be, because as you've probably seen, there are toplevel-style things that set this.

For the `WM_TRANSIENT_FOR`, I can't recall now but I thought I remember seeing stuff that doesn't make sense as a popup setting it. I guess it might be fine with dialogs, but I'm not positive, might need some extra testing.

For `_NET_WM_STATE_ABOVE`, I'm also not confident in this being a good popup indicator because I can imagine normal windows setting this erroneously, but I'm not positive actually, and yabridge doesn't seem to be giving us much to work with in terms of heuristics, lol.

I really should've documented some of this as I came across, and that was really some of the intention behind the integration tests, but clearly it wasn't enough...

Unk-now:
Hello @En-En-Code and @Supreeeme,
Thank you for your work on the recent popup-heuristics-fixes!
I've built the popup-heuristics-fixes branch (27bbedd) and tested it with Clip Studio Paint on Wine. I can confirm that the displaced menu issue seems to be resolved on my end. The menus now appear in the expected position, directly below the parent menu item.
This is a significant improvement and I'm very hopeful this fix will be adopted. I understand there might be ongoing discussions or considerations regarding potential side effects, but from a user perspective, this change has resolved the specific problem I was encountering.
Thank you again for your efforts on this!

GoranKovac:
Yeah it seems that incorporating `yabridge bridged` specific fixes breaks normal apps. For example https://github.com/Supreeeme/xwayland-satellite/issues/291 this installer is broken in latest stable but was fixed with https://github.com/Supreeeme/xwayland-satellite/pull/298/commits/f1841e0057d05f5cdfa93a733ad4676f1faba5be but that then broke reaper (and probably other apps). Now with latest commits in this PR installer autocloses when it gets unfocused.
Also some modal windows use `_NET_WM_STATE_ABOVE` https://github.com/Supreeeme/xwayland-satellite/issues/293#issuecomment-3608552551

Not sure if best solution but would it make sense to isolate `yabridge` by class (probably not)?
Anyway since my whole library yabridged plugins is I will document `xprops` and `xwininfo` but it seems that the most (not all) problematic plugins are the ones made with `JUCE` framework https://juce.com/.

GoranKovac:
[satelite_yabridge_doc.log](https://github.com/user-attachments/files/23931484/satelite_yabridge_doc.log)


So here is detailed log with `xprop` and `xwininfo` for problematic plugins. Note that all of them suffer from hang/freeze which last for about 10 sends when trying to activate menu item or close menu entierly (this is also for plugins that have tooltip issues).

If you want me to move this into discussion or another BR let me know (I wont post here further)

En-En-Code:
> I feel like SKIP_TASKBAR should really only be used as a last resort if we can't figure out what kind of window it should be

I agree. That codepath was likely more used before, but since then, `MENU`, `TOOLTIP`, `DND`, `POPUP_MENU`, `DROPDOWN_MENU` and `UTILITY` all have special handling (and I also have a solid idea of how `SPLASH` should be handled), so there are a lot fewer `_NET_WM_WINDOW_TYPE`s which would end up down there.

> For the WM_TRANSIENT_FOR, I can't recall now but I thought I remember seeing stuff that doesn't make sense as a popup setting it.

[You have mentioned that before](https://github.com/Supreeeme/xwayland-satellite/issues/110#issuecomment-2745826682). I can see how the [wording of `WM_TRANSIENT_FOR` in the ICCCM](https://tronche.com/gui/x/icccm/sec-4.html#s-4.1.2.6) could be read to imply the window is still toplevel. That is why I have used it as a heuristic for only window types where it resolves reported issues. I'm trying not to worry about hypotheticals.

> For _NET_WM_STATE_ABOVE, I'm also not confident in this being a good popup indicator

I am not sure about how I feel about it either. I did note in the relevant issue [its potential for edge cases](https://github.com/Supreeeme/xwayland-satellite/issues/293#issuecomment-3607638447) and at least one such case already arose. Deciding the exact heuristics is part of why I moved it to draft. I am expecting some compromise to be inevitable.

> I really should've documented some of this as I came across, and that was really some of the intention behind the integration tests, but clearly it wasn't enough...

In a way, tests are documentation of a system's behavior. The explanations for why we expect the result we do and the issues to reference back at make it easier to understand and verify. Considering `popup_heuristics` takes 1.6 seconds to run in its current state (lol), I want to rewrite `guess_is_popup` to make it unit-testable, and I have ideas for how I would do that.

GoranKovac:
Wanted to give one more try with `yabridge`:
https://github.com/GoranKovac/xwayland-satellite/tree/yabridge_popups

After extensive testing it seems when the `WM_CLASS` is `yabridge-host.exe.so` its a popup. Tested across large library and works really well. I have not yet encountered any plugin that breaks or misbehaves in sense that it spawns as popup. 
Superior drummer works, Slate Drums and all other Slate Digital plugins work, all Valhallas, TSE and many many more.

The only issue I have found is with `A.O.M` plugins that does not spawn popup on specific button even the popup also has same `WM_CLASS` and same `xprop` as any working plugin:

https://github.com/user-attachments/assets/d5e6ff49-73e5-436a-9ca1-b479ceaf1b76

```
WM_STATE(WM_STATE):
		window state: Normal
		icon window: 0x0
_NET_WM_ICON(CARDINAL) = 	Icon (32 x 32):

_NET_WM_STATE(ATOM) = _KDE_NET_WM_STATE_SKIP_SWITCHER, _NET_WM_STATE_ABOVE, _NET_WM_STATE_SKIP_PAGER, _NET_WM_STATE_SKIP_TASKBAR
_NET_WM_NAME(UTF8_STRING) = "menu"
WM_ICON_NAME(STRING) = "menu"
WM_NAME(STRING) = "menu"
WM_HINTS(WM_HINTS):
		Client accepts input or input focus: False
		Initial state is Normal State.
		bitmap id # to use for icon: 0x6003a5
		bitmap id # of mask for icon: 0x6003a7
		window id # of group leader: 0x800022
_NET_WM_WINDOW_TYPE(ATOM) = _NET_WM_WINDOW_TYPE_NORMAL
_MOTIF_WM_HINTS(_MOTIF_WM_HINTS) = 0x3, 0x24, 0x0, 0x0, 0x0
		program specified location: 1952, 112
		program specified minimum size: 273 by 246
		program specified maximum size: 273 by 246
		window gravity: Static
_NET_WM_USER_TIME_WINDOW(WINDOW): window id # 0x600014
XdndAware(ATOM) = BITMAP
_NET_WM_PID(CARDINAL) = 39330
WM_LOCALE_NAME(STRING) = "en_US.UTF-8"
WM_CLIENT_MACHINE(STRING) = "archlinux"
WM_CLASS(STRING) = "yabridge-host.exe.so", "yabridge-host.exe.so"
WM_PROTOCOLS(ATOM): protocols  WM_DELETE_WINDOW, _NET_WM_PING, WM_TAKE_FOCUS
```

Here is the log when clicking on it. Really not sure whats different about it than the rest.
[satellite.log](https://github.com/user-attachments/files/24060193/satellite.log)

Sometimes (very very rarely when clicking like 200 times) it spawns but far away from the plugin window.

EDIT: It seems its not spawning because it creates 1920 px offsetted from main window, if main VST is moved then it can spawn more often:

image

EDIT2:

Hm..... everything is correct and consistent (no more random spawn positions on ANY plugin mentioned at the end of this post https://github.com/Supreeeme/xwayland-satellite/issues/302#issuecomment-3623099382 ) when this is commented out?

```
SurfaceRole::Popup(Some(popup)) => {
                // popup.positioner.set_offset(
                //     ((event.x() as i32 - win.output_offset.x) as f64 / scale_factor.0) as i32,
                //     ((event.y() as i32 - win.output_offset.y) as f64 / scale_factor.0) as i32,
                // );
                // popup.positioner.set_size(
                //     1.max((event.width() as f64 / scale_factor.0) as i32),
                //     1.max((event.height() as f64 / scale_factor.0) as i32),
                // );
                // popup.popup.reposition(&popup.positioner, 0);
            }
```

https://github.com/user-attachments/assets/6099e085-0c40-49c1-bf5e-1d956684eeae

En-En-Code:
Ok, because I've noticed #323 trying to solve some of the same issues, I'll try to resolve some of the open questions here.

First, after some research, `_NET_WM_STATE_ABOVE` seems to mostly be applied to `_NET_WM_WINDOW_TYPE_DOCK`, `NOTIFICATION`, `TOOLBAR`, etc., which satellite currently is not trying to handle correctly anyways. It was mostly a fix for Yabridge, which has proven to be more effectively marked as an exception to the rules (see #315). I do not mind removing it as a check.

`WM_TRANSIENT_FOR` is also a complicated case. Recently someone finally found [a window with `TRANSIENT_FOR` which needs to be a top-level](https://github.com/Supreeeme/xwayland-satellite/issues/294#issuecomment-3650978286) (it's even marked with the `UTILITY` type). That said there are also open issues where recognizing `TRANSIENT_FOR` on a normal (#291) and dialog (#280) would resolve them. `TRANSIENT_FOR` does contain the window it is parented to (or the special case of none/root), but I am not sure how much information that actually gives.

Honestly, the most likely scenario is for this to be rejected and I'll come back with a PR which just fixes the inaccuracies of `popup_heuristics`, documents them better, and possibly refactors `guess_is_popup` to make them unit tests instead of integration tests.

En-En-Code:
With #323 merged, this branch is no longer needed. I would still like to add good documentation to the popup tests and refactor `guess_is_popup`, but that should go on a clean branch.

Issue #302, opened 2025-12-05
Some WINE windows (menus) take a long time to delete


GoranKovac:
### Operating System

Arch

### Wayland Compositor

Hyprland

### Describe the issue.

Moving this into separate thread for visibility since it seems another problem. 
Previously reported here https://github.com/Supreeeme/xwayland-satellite/issues/293#issuecomment-3613900895 and https://github.com/Supreeeme/xwayland-satellite/issues/293#issuecomment-3614524757

What seem to happen is something is hanging/take a lot of time to destroy windows. While these windows are "special" meaning they are some kind of decorations (dropshadows from JUCE framework). There are 4 of them created around the main window (menu)

I was not able to trace down in source code where it happens (tried guess working) but DEBUG log always works slow when doing this:

Copy paste from previous report:
```
2025-12-04T19:07:53.365Z DEBUG xwayland_satellite::xstate            > changed property "_NET_WM_STATE" for Window { res_id: 8388619 }
 2025-12-04T19:07:53.365Z DEBUG xwayland_satellite::xstate            > Window { res_id: 8388619 } request: X | Y | WIDTH | HEIGHT
 2025-12-04T19:07:53.365Z DEBUG xwayland_satellite::server::dispatch  > deleting (surface 54)
 2025-12-04T19:07:53.365Z DEBUG xwayland_satellite::xstate            > destroying window Window { res_id: 8388619 }
 2025-12-04T19:07:55.366Z DEBUG xwayland_satellite::server            > unmapping Window { res_id: 8388621 }
 2025-12-04T19:07:55.366Z DEBUG xwayland_satellite::xstate            > changed property "_NET_WM_STATE" for Window { res_id: 8388621 }
 2025-12-04T19:07:55.366Z DEBUG xwayland_satellite::xstate            > Window { res_id: 8388621 } request: X | Y | WIDTH | HEIGHT
 2025-12-04T19:07:55.366Z DEBUG xwayland_satellite::server::dispatch  > deleting (surface 57)
 2025-12-04T19:07:55.366Z DEBUG xwayland_satellite::xstate            > destroying window Window { res_id: 8388621 }
 2025-12-04T19:07:57.366Z DEBUG xwayland_satellite::server            > unmapping Window { res_id: 8388623 }
 2025-12-04T19:07:57.367Z DEBUG xwayland_satellite::xstate            > changed property "_NET_WM_STATE" for Window { res_id: 8388623 }
[...]

PR #303, opened 2025-12-07
Fix wine windows hanging

GoranKovac:
Some wine apps expect WM_STATE to be updated when mapping/unmapping windows. Unless handled properly wine app would timeout while waiting for update state

Fixes https://github.com/Supreeeme/xwayland-satellite/issues/302

I did not see any regressions, but to fully utilize this `yabridge` plugins would need to spawn menus as popups.

### DIFF
--- a/src/server/event.rs
+++ b/src/server/event.rs
@@ -934,16 +934,20 @@ impl Event for client::wl_touch::Event {
             } => {
                 let mut cmd = CommandBuffer::new();
                 {
+                    let connection = &mut state.connection;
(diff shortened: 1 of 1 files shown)


Supreeeme:
Also be sure to run `cargo fmt`.

Supreeeme:
That did not do what I thought it would :sweat_smile:

Supreeeme:
Okay I just went ahead and pushed your changes to main, sorry about that. Thanks for the quick fix!

Issue #221, opened 2025-08-20
REAPER + ValhallaPlate dropdowns are not treated as popups


slipinthedove:
### Operating System

Arch Linux (running under linux-cachyos-6.16.1-2, but also applicable on the regular kernel)

### Wayland Compositor

niri (commit YaLTeR/niri@1013147ba3154788cb16ffc32b8155fd2008dfa3, but also applicable on latest stable version. as of writing - v25.01.1)

### Describe the issue.

Certain custom dropdown menus (namely, audio plugins) when loading them are snatched to spawn at the center regardless of the cursor/window location.
Below is an example of REAPER (XWayland) running through the ValhallaPlate (Wine 10.13 container) as i open a dropdown to select a mode and such appearing on the center of my workspace rather than the desired/optimal location.

<img width="865" height="460" alt="Image" src="https://github.com/user-attachments/assets/1050a7d9-dc2d-4796-8fbb-367393f50079" />

This is not a wine-specific thing, either, since there may be some people that questions this out. 

Running native applications that have to run under XWayland (thanks to the DAW nor the graphical library of such not supporting another framework) also creates the same behavior. Below is an example running the latest version of MiniMeters and then trying to open the hamburger menu from such, once again with the same centered-window behavior.

<img width="747" height="513" alt="Image" src="https://github.com/user-attachments/assets/ec8018ef-ad3c-4090-9d38-a374d3763ca9" />

I am not sure if this is a mis-doing of my own or an actual bug from xwayland-satellite itself, but regardless, i would love to see how this unfolds as this has been bothering me even as far back as 0.6.

### Steps to reproduce

Open a dropdown (ideally inside a DAW, but could also be applied on standalone) of an audio-based plugin through XWayland.

### Please confirm you have reproduced this on the latest commit.

- [x] I have reproduced this on the latest commit

### Below is a debug log with me opening REAPER, creating a new track, opening ValhallaPlate, and then opening the respective dropdown.

[xwl.log](https://github.com/user-attachments/files/21903022/xwl.log)

Supreeeme:
The primary issue here is that satellite is not properly marking these drop downs as popups, but instead as normal toplevels. This is purely heuristics based so it may be that these windows just cannot be properly marked as popups without sacrificing other window types. But, I can take a look. 
Can you run `xprop` on the drop down menus and post the output here?

slipinthedove:
> Can you run `xprop` on the drop down menus and post the output here?

[...]

Issue #308, opened 2025-12-10
Misbehaviour in reconfiguring_window for WINE popups


GoranKovac:
### Operating System

Arch

### Wayland Compositor

Hyprland

### Describe the issue.

Sometimes popups appear where they should be but sometimes they just spawn offset or not at all (also mentioned here https://github.com/Supreeeme/xwayland-satellite/pull/298#issuecomment-3631867840 ) . After investigating I've found its happening here:

https://github.com/Supreeeme/xwayland-satellite/blob/f0ad674b7009a6afd80cea59d4fbf975dd68ee95/src/server/mod.rs#L1000

If this whole section is commented out then WINE popups work fine all the time but breaks rest of the popups and drag & drop.

This can be tested with `Yabridge` plugins on my current branch:
https://github.com/GoranKovac/xwayland-satellite/tree/yabridge_popups

At some random time when opening menus it spawns far away

https://github.com/user-attachments/assets/a929c971-9477-4e08-a0d3-da25111df0ce

Same happens here but also one dropdown (top left button) does not spawn at all in very very rare case after spamming click for 300 times it appears far far offset.

https://github.com/user-attachments/assets/d9ba5e7f-984a-44c1-b461-563533561db9

When `SurfaceRole::Popup` body is commented out then everything works fine 100% of the time:

https://github.com/user-attachments/assets/7fcad268-84e1-4fed-b5ea-16189412880e

Did not recorded spamming menu in superior drummer since offset spawn just does not happen anymore now (but tested it for a long time).

Log for superior drummer (happened to offset immediately  on first menu spawn)

[satellite.log](https://github.com/user-attachments/files/24077500/satellite.log)

A.O.M Invisible limiter not spawning top left menu log:

[satellite.log](https://github.com/user-attachments/files/24077533/satellite.log)

### Steps to reproduce

`Requires Yabridge setup`
1. Open Reaper
2. Add TSE X50 https://www.tseaudio.com/software/x50v2 (Or valhalla plugins)
3. Start spamming some menu or right click context

Result: Eventually they will spawn offset, just a matter of time

### Please confirm you have reproduced this on the latest commit.

- [x] I have reproduced this on the latest commit

GoranKovac:
```
[...]

PR #315, opened 2025-12-12
Detect yabridge popups (hardcoded wm_class)

GoranKovac:
Anything that has "yabridge-host" as a `wm_class` is 100% popup, tooltip, drag and drop operation. Yabridge popups cant be properly identified with ATOMS.

Yabridge popups come in 4 forms:
1. Tooltips
2. Menu popups
3. Right Click context menus
4. Drag and drop within the plugin

All of them have same generic `xprop` ATOMS such as: `_KDE_NET_WM_STATE_SKIP_SWITCHER, _NET_WM_STATE_ABOVE, _NET_WM_STATE_SKIP_PAGER, _NET_WM_STATE_SKIP_TASKBAR` and its always `_NET_WM_WINDOW_TYPE(ATOM) = _NET_WM_WINDOW_TYPE_NORMAL`. This makes it difficult to differentiate them from other popups and introduce regressions. One thing in common they all have is that `WM_CLASS = "yabridge-host.exe`. This is more of a hack but it isolates `yabridge` from rest of them.

Predefined array of known popup wm_classes is used here for potentially expanding it in future for other apps.

Fixes: https://github.com/Supreeeme/xwayland-satellite/issues/221 , https://github.com/Supreeeme/xwayland-satellite/issues/293

After thoroughly testing this for a while now in `Reaper` I can confirm it works without any regressions or misbehaviors. Huge number of plugins now work properly with popups, tooltips and drag and drop.

Not sure how you feel about this fix since its a hardcoded workaround but currently don't seem to find any other better solution. Feel free to close it if you don't like this approach.

### DIFF
--- a/src/xstate/mod.rs
+++ b/src/xstate/mod.rs
@@ -697,7 +697,7 @@ impl XState {
         wm_class: Option<String>,
     ) -> XResult<bool> {
         if let Some(class) = wm_class {
-            // If class is predefined is popup
+            // If class is predefined it's a popup
             if KNOWN_POPUP_WMCLASS.contains(&class.as_ref()) {
                 return Ok(true);
             }


GoranKovac:
Additionally tested with:
`Ardour`, `Bitwig Demo`, `Carla`, `Tracktion-waveform`

Everything works fine here too.

Supreeeme:
I don't love hardcoding this because any random update can break it. But if there's no other way it is what it is. I need to think about it/look into it a bit more myself first though.

GoranKovac:
> I don't love hardcoding this because any random update can break it. But if there's no other way it is what it is. I need to think about it/look into it a bit more myself first though.

Just so you are aware if you decide to merge this there is additional work needed for them mentioned here https://github.com/Supreeeme/xwayland-satellite/issues/308 . Maybe its a separate issue maybe not but just so you are aware.

GoranKovac:
Just a small observation, when plugins are added it spawns individual `yabridge` process per plugin and their popups point to corresponding `PID`. Not sure is the `WM_CLASS` extracted from that process or what but wanted to share this if it might bring some clarity:
image

GoranKovac:
I have found a way to reliably detect these types of popups without hard coding but it depends on https://github.com/Supreeeme/xwayland-satellite/pull/323 (decorations refactor) so will create PR once it gets merged.

9Prestidigitator:
Omg this fixes nearly every issue I've had with audio software UI's on Niri/xwayland. Thank you so much, great work!

Issue #277, opened 2025-11-12
The WeChat search box fails to pop up, and the right-click pop-up menu for messages also doesn't display.


falser101:
### Operating System

ArchLinux

### Wayland Compositor

Niri

### Describe the issue.

The WeChat search box fails to pop up, and the right-click pop-up menu for messages also doesn't display.

### Please confirm you have reproduced this on the latest commit.

- [x] I have reproduced this on the latest commit

Supreeeme:
Could you provide logs? Run satellite with `RUST_LOG=debug` set and try to open the menus.

tt66ea:
<img width="1280" height="492" alt="Image" src="https://github.com/user-attachments/assets/283693f5-2299-4ab3-bc5c-6f7d709f4e9f" />
seems like that the right click menu display in a wrong place btw

and the menu doesn't work because it will disappear when I try to move the cursor on it

Wayland Compositor: Niri

tt66ea:
[xwl-wx.log](https://github.com/user-attachments/files/23627181/xwl-wx.log)

run wechat-universal under niri and here's the log

Supreeeme:
I think the right click menu thing is a dupe of #265, which was just fixed. Is the search box not appearing still an issue?

GoodbyeNJN:
> I think the right click menu thing is a dupe of [#265](https://github.com/Supreeeme/xwayland-satellite/issues/265), which was just fixed. Is the search box not appearing still an issue?

The issue with the search popup not appearing still exists, but it can be resolved through some workarounds, such as setting window rules in niri like so:

```
window-rule {
    match title=r#"^wechat$"# app-id=r#"^wechat$"#
    open-focused false
}
```

It appears the search popup logic is to show when focus is on the search box and close when focus is lost. By setting to not auto-focus, you can avoid losing focus from the input box.

So in a sense, perhaps this should not be considered an issue of xwayland-satellite, but the issue of the search popup (as well as emoji, main menu popups, etc.) appearing in the wrong position may still require further investigation.

Supreeeme:
It sounds like the popup is being created as a toplevel when it probably shouldn't? Can you send the output of `xprop` on the popup, and also provide a screenshot?

GoodbyeNJN:
> It sounds like the popup is being created as a toplevel when it probably shouldn't? Can you send the output of `xprop` on the popup, and also provide a screenshot?


From the search pop-up window:

```
WM_STATE(WM_STATE):
[...]

PR #323, opened 2025-12-14
Detecting if UTILITY is popup

GoranKovac:
Ardour uses UTILITY atom all over the place for toplevel windows: Plugins, Dialogs etc
However it does not provide any MOTIF_HINTS at all.

WeChat however uses them for popups and also provides MOTIF_HINTS with flags 0x2 indicating that only decorations are active. MaterialMaker also follows what WeChat is doing for right click menus.

This fix assigns UTILITY as popup ONLY if MOTIF_HINTS are provided and functions are not active (first bit is not set).

Fixes: https://github.com/Supreeeme/xwayland-satellite/issues/294
Does not regress (per user testing https://github.com/Supreeeme/xwayland-satellite/issues/294#issuecomment-3651756248 ): https://github.com/Supreeeme/xwayland-satellite/issues/277

Also some other apps (MaterialMaker) works fine with right click context menus.

### DIFF
--- a/src/xstate/mod.rs
+++ b/src/xstate/mod.rs
@@ -689,8 +689,7 @@ impl XState {
         let mut motif_popup = false;
         if let Some(hints) = motif_hints {
             // If MOTIF_WM_HINTS provides no decorations for client assume its a popup
-            let is_client_decoration = hints.decorations.is_some_and(|d| d.is_clientside());
-            motif_popup = is_client_decoration && hints.decorations.is_some_and(|d| d.is_empty());
+            motif_popup = hints.decorations.is_some_and(|d| d.is_clientside());
             // If the motif hints indicate the user shouldn't be able to do anything
             // to the window at all, it stands to reason it's probably a popup.
             if hints.functions.is_some_and(|f| f.is_empty()) {
@@ -1159,7 +1158,7 @@ mod motif {
 
     bitflags! {
         #[derive(Debug, PartialEq, Eq, Clone, Copy)]
-            pub struct Decorations: u32 {
+        pub struct Decorations: u32 {
             const All = 1;
             const Border = 2;
             const Resizeh = 4;
@@ -1199,14 +1198,6 @@ mod motif {
 
         pub fn is_serverside(&self) -> bool {
             !self.is_empty()
-        }
-
-        pub fn clientside(&self) -> zxdg_toplevel_decoration_v1::Mode {
-            zxdg_toplevel_decoration_v1::Mode::ClientSide
-        }
-
-        pub fn serverside(&self) -> zxdg_toplevel_decoration_v1::Mode {
-            zxdg_toplevel_decoration_v1::Mode::ServerSide
         }
     }
 
--- a/tests/integration.rs
+++ b/tests/integration.rs
@@ -2041,7 +2041,7 @@ fn popup_heuristics() {
     );
     f.map_as_popup(&mut connection, git_gui_dropdown);
 
-    let wechat_popup = connection.new_window(connection.root, 10, 10, 50, 50, false);
+    let wechat_popup = connection.new_window(connection.root, 10, 10, 50, 50, true);
     connection.set_property(
         wechat_popup,
         x::ATOM_ATOM,


src/xstate/mod.rs
```diff
@@ -686,7 +686,10 @@ impl XState {
         motif_hints: Option<motif::Hints>,
         has_transient_for: bool,
     ) -> XResult<bool> {
+        let mut no_function_motif = false;
         if let Some(hints) = motif_hints {
+            // indicating motif flags first bit is not set (eg. 0x2 decorations only)
+            no_function_motif = hints.functions.is_none();
```
Supreeeme:
```suggestion
            no_function_motif = hints.functions.is_none_or(|f| f.is_empty());
```
In the name of future proofing.

Supreeeme (commented):
Looks like the `popup_heuristics` test needs to be updated. Also, there should be a test case for this as well.

GoranKovac:
> Looks like the `popup_heuristics` test needs to be updated. Also, there should be a test case for this as well.

Yeah was not sure how test should look since it mixes atoms with motif_hints

```
connection.set_property(
        wechat_popup,
        x::ATOM_ATOM,
        connection.atoms.win_type,
        &[connection.atoms.win_type_utility],
);
connection.set_property(
        wechat_popup,
        x::ATOM_ATOM,
        connection.atoms.motif_wm_hints,
        &[0x2_u32, 0, 0, 0, 0],
);
connection.set_property(
        wechat_popup,
        x::ATOM_ATOM,
        connection.atoms.motif_wm_hints,
        &[0x3_u32, 0, 0, 0, 0],
);

```
Like this?

GoranKovac:
Updated and added all tests for wechat and ardour. Both pass correctly.

EDIT: Sorry, removed ardour test (its not a popup)

GoranKovac:
Spotted that `Godot` and some Godot based apps `PixelOver` marks their popup (`_NET_WM_WINDOW_TYPE_UTILITY`) with override redirect so pushed a fix for that.
Tested with rest apps (Ardour, MaterialMaker, Godot....) no regressions

src/xstate/mod.rs
```diff
@@ -686,7 +686,10 @@ impl XState {
         motif_hints: Option<motif::Hints>,
         has_transient_for: bool,
     ) -> XResult<bool> {
+        let mut no_function_motif = false;
         if let Some(hints) = motif_hints {
+            // indicating motif flags first bit is not set (eg. 0x2 decorations only)
+            no_function_motif = hints.functions.as_ref().is_none_or(|f| f.is_empty());
```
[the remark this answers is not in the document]
En-En-Code: My bad. I read this incorrectly. `hints` is an `Option` but `hints.functions` inside of `motif::Hints` is *also* an `Option`, which I missed and this misinterpreted the code. I do not think I understand why `hints.functions` needs to be an `Option`, but at least makes sense and is fine by me.

tests/integration.rs
```diff
@@ -2048,6 +2048,18 @@ fn popup_heuristics() {
...
+        connection.atoms.motif_wm_hints,
+        connection.atoms.motif_wm_hints,
+        &[0x2_u32, 0, 0, 0, 0],
+    );
+    connection.set_property(
+        wechat_popup,
+        connection.atoms.motif_wm_hints,
+        connection.atoms.motif_wm_hints,
+        &[0x3_u32, 0, 0, 0, 0],
+    );
```
Supreeeme: Why is this set twice?

GoranKovac:
> The Ardour test would actually still be useful for verifying that not all utility windows end up as popups.

Where should that test be implemented?

Also will need to remove override_redirect from `wechat` test. Would you prefer to add another test for `godot` which uses same type but its flagged with override redirect?

```
let godot_popup = connection.new_window(connection.root, 10, 10, 50, 50, true);
    connection.set_property(
        godot_popup,
        x::ATOM_ATOM,
        connection.atoms.win_type,
        &[connection.atoms.win_type_utility],
    );
    connection.set_property(
        godot_popup,
        connection.atoms.motif_wm_hints,
        connection.atoms.motif_wm_hints,
        &[0x2_u32, 0, 0, 0, 0],
    );
```

These are all findings from xprops above (https://github.com/Supreeeme/xwayland-satellite/pull/323#issuecomment-3657798271 https://github.com/Supreeeme/xwayland-satellite/pull/323/commits/3de808b1ed460ad3fb21df73a331f2f5a05b2cdd)

En-En-Code:
> Where should that test be implemented?

In `popup_heuristics`.

> Also will need to remove override_redirect from wechat test. Would you prefer to add another test for godot which uses same type but its flagged with override redirect?

You could. You could instead write a test with a purely hypothetical utility window with `override_redirect` and no `_MOTIF_WM_HINTS` (I did a quick Internet search but could not find any `xprop` suggesting this kind of window) since this would cover the missed edge case better.

GoranKovac:
@En-En-Code  Yes exactly:

Having `MOTIF_WM_HINTS` but no function bitflag is the main difference is it a popup or toplevel. In case of `Ardour` it completely lacks of `MOTIF_WM_HINTS` so it considered to be toplevel.

Sure checking `decorations` would work too I guess but I was under impression based on my research and https://github.com/Supreeeme/xwayland-satellite/blob/1bbd5cae7ec80241e8b44629d4da503ed7cf18c9/src/xstate/mod.rs#L692
that `functions` are the main thing that defines is it a popup.

Also regarding `Ardour` it does popups in its own windows (not exactly sure whats the right terminology since this x11 is very new to me). When they are spawned they grab cursor and they cant be queried with `xprop` (`xprop: error: Can't grab the mouse.`) so I don't really know what type they are exactly. But 90% of the toplevel windows are `_NET_WM_WINDOW_TYPE_UTILITY` and lack of `MOTIF_WM_HINTS`

En-En-Code:
> Sure checking decorations would work too I guess but I was under impression ... that functions are the main thing that defines is it a popup.

The reason `functions` from `_MOTIF_WM_HINTS` was being checked originally was for if the window explicitly requested the user to *not* be able to perform any resize, move, close, etc. functions on the window. That check is for window which behave like tooltips but do not set `_NET_WM_WINDOW_TYPE_TOOLTIP`. Utility windows, on the other hand, are designed to have user interaction, so I would not except `functions` to apply.

I was interested in using the request to have no decorations as the hint because top-levels can have title bars, window borders, etc. while pop-ups (generally) do not, so it made more sense within my mental model. That's why I suggested it when I saw it would work the same across all your provided `xprop` outputs.

GoranKovac:
> > Sure checking decorations would work too I guess but I was under impression ... that functions are the main thing that defines is it a popup.
> 
> The reason `functions` from `_MOTIF_WM_HINTS` was being checked originally was for if the window explicitly requested the user to _not_ be able to perform any resize, move, close, etc. functions on the window. That check is for window which behave like tooltips but do not set `_NET_WM_WINDOW_TYPE_TOOLTIP`. Utility windows, on the other hand, are designed to have user interaction, so I would not except `functions` to apply.
> 
> I was interested in using the request to have no decorations as the hint because top-levels can have title bars, window borders, etc. while pop-ups (generally) do not, so it made more sense within my mental model. That's why I suggested it when I saw it would work the same across all your provided `xprop` outputs.

Decorations struct would need to be expanded to check its flags (like functions) in that case. Since currently it only shows are decorations server or client side and pass that info to wayland.
Or we can check needed info just based on are which decorations type it is?

EDIT:
Is this better/preferable https://github.com/GoranKovac/xwayland-satellite/tree/utility-fix-decorations ?

1. Added decoration bitflags
2. Check if decoration is client sided & are decorations flags not set

All tests pass, works same as current branch, no regressions seen.

If so will merge into this branch.

En-En-Code:
> Decorations struct would need to be expanded to check its flags (like functions) in that case. Since currently it only shows are decorations server or client side and pass that info to wayland.
Or we can check needed info just based on are which decorations type it is?

At the moment, you could just check if clientside decorations are set, but I think the existing code here is incorrect, so I somewhat like that you did not. Anything with a decorations hint that is not 0 or 1 has no decorations at all (even though a subset of decorations was selected), and the value of 0, which represents no decorations, sets clientside decorations which draw a titlebar, even though the titlebar bit is unset. However, this issue is outside of the scope of this PR.

I am good with the changes on `utility_fix_decorations`.

GoranKovac:
> > Decorations struct would need to be expanded to check its flags (like functions) in that case. Since currently it only shows are decorations server or client side and pass that info to wayland.
> > Or we can check needed info just based on are which decorations type it is?
> 
> At the moment, you could just check if clientside decorations are set, but I think the existing code here is incorrect, so I somewhat like that you did not. Anything with a decorations hint that is not 0 or 1 has no decorations at all (even though a subset of decorations was selected), and the value of 0, which represents no decorations, sets clientside decorations which draw a titlebar, even though the titlebar bit is unset. However, this issue is outside of the scope of this PR.
> 
> I am good with the changes on `utility_fix_decorations`.

Ok, how do you suggest moving this forward? Merge? Squash? Open new clean PR?

src/xstate/mod.rs
```diff
@@ -1151,10 +1160,23 @@ mod motif {
...
+            const Minimize = 32;
+            const Maximize = 64;
+        }
+    }
+
     #[derive(Default)]
     pub(super) struct Hints {
         pub(super) functions: Option<Functions>,
         pub(super) decorations: Option<Decorations>,
+        pub(super) decorations_hints: Option<DecorationsHints>,
```
Supreeeme: This is kind of confusing imo, the struct kind of looks like it would be getting `decorations` and `decorations_hints` from two different sources but it's the same one. Instead I would suggest:
- replace `Decorations` with `DecorationsHints` (but still call it `Decorations`)
- add some kind of helper function, like `Decorations::is_clientside()` in place of wherever `Decorations::Client` is currently used

Supreeeme: Yeah, this would be refactoring `Decorations` everywhere, but it's not used in that many places, and should be straightforward to change:
- Create helper  functions on `Decorations`: `fn is_clientside(&self) -> bool` and `fn clientside() -> Self` (and do the same for serverside)
- Anywhere that `Decorations::Client` is checked/compared against, replace it with `Decorations::is_clientside` (and do the same for serverside)
- Anywhere that `Decorations::Client` is actually created, replace it with `Decorations::clientside()`

Probably a little more than you bargained for but I think it's relatively straightforward. Let me know if you have any issues.

Supreeeme:
Thanks a lot!

PR #327, opened 2025-12-20
optionally load decoration font at runtime instead of embedding it

### LINKED ISSUE #317: RFE: option to resolve and load font for window decorations at runtime
**Is your feature request related to a problem? Please describe.**

I'm packaging xwayland-satellite for Fedora Linux (primarily for use with niri). I noticed that v0.8 now includes a bundled font (Open Sans Regular) - which is not a problem in itself, but against our distribution policies (I would need to replace the font with one that we distribute).

**Describe the solution you'd like**

I think it would be nice if xwayland-satellite could (optionally?) use fontconfig to resolve the font it needs (it could maybe even be a configurable font!) and load that at runtime, instead of statically embedding Open Sans Regular into the executable.

**Describe alternatives you've considered**

Option 1: I just replace the Open Sans Regular TTF file with one that is distributed by Fedora, this does not require code changes.
Option 2: Patch xwayland-satellite to use fontconfig to look up the font and load it at runtime instead of embedding it in the executable.
Option 3: Do both, so the embedded font can work as a fallback (not my preferred option, since it still has the problem of "I need to replace the font that is embedded during the build").

**Additional context**

Looking at the API docs for the `fontconfig` crate, this should be relatively straightforward to implement. If you are open to this idea in general, I can work on implementing this and submitting a PR for it.

decathorpe:
Fixes #317

I moved the font data to a separate `static` - otherwise the `FontRef` in the dynamic code path is a dangling reference (since the `Vec<u8>` loaded from disk is not `&'static` as opposed to the return type of `include_bytes!()`).

The original code path also panic!s if the font could not be loaded, so I think it's reasonable to have the new code path panic! on failure too.

I tested this patch on my system and it appears to work correctly.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -234,6 +234,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "dlib"
+version = "0.5.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "330c60081dcc4c72131f8eb70510f1ac07223e5d4163db481a04a0befcffa412"
+dependencies = [
+ "libloading",
+]
+
+[[package]]
 name = "downcast-rs"
 version = "1.2.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


decathorpe:
(force-pushed after running `cargo fmt` to make the CI happy)

decathorpe:
Thanks for taking a look so quickly! :)

Issue #335, opened 2025-12-28
Forward X options to Xwayland


mviereck:
**Is your feature request related to a problem? Please describe.**
I am integrating xwayland-satellite in x11docker. 
x11docker sets several X options, most noteable `-auth file`, further those like `-extension MIT-SHM +extension SECURITY`.
This currently does not work with xwayland-satellite, it throws an error.
Because of missing option `-auth` x11docker cannot set up a cookie authentication. Other options are less important, but desired, too, e.g. `-listen tcp`.

**Describe the solution you'd like**
Please forward X options to Xwayland.

**Additional context**
`Xwayland -h` shows more options than `man Xwayland`. Not all of them might be supported, but seemingly more than documented in the manpage.

En-En-Code:
I have actually wanted to get around to this.

`satellite` should not support every `Xserver`/`Xwayland` command-line option. Some flags conflict with flags we set (e.g. all of the flags incompatable with `--rootless`), some do not make sense to set, and some could break functionality, which is the problem I am mst concerned about, as it would be harder to get correct just by reading through documentation. In theory I can write a script that adds each option and trials it against the integration tests, which would be a good starting list.

Based on some initial research, `auth` or `listen` should not be a problem, but the `SECURITY` extension does change selection mechanism and may also break scaling since it relies in `XRandR`.

mviereck:
> satellite should not support every Xserver/Xwayland command-line option. Some flags conflict with flags we set (e.g. all of the flags incompatable with --rootless), some do not make sense to set, and some could break functionality, which is the problem I am mst concerned about, as it would be harder to get correct just by reading through documentation. 

ok, I see.

>  In theory I can write a script that adds each option and trials it against the integration tests, which would be a good starting list.

This might be to much work for only rare use cases.

> Based on some initial research, auth or listen should not be a problem

That sounds good.

In general I see three possible approaches:
 - Check all options with a trial script as you said. Likely too much work, as manual re-checks are needed, and only few people need further options.
 - Provide an option for custom Xwayland option settings without any check by xwayland-satellite. It's up to the user to check if it works for him and allows to break functionality. A warning in documentation would be enough.
 - Allow only a subset of useful non-conflicting options.

I had a look which options are really important for x11docker:
 - `-auth file` for cookie setup
 - `-extension MIT-SHM` to disable shared memory.

Nice to have:
 - `+extension SECURITY` for trusted or untrusted cookies (fallback to regular cookies available)
[...]

Issue #75, opened 2024-11-18
how to disable listening on abstract socket?


odomingao:
Hi,

`xwayland-satellite -nolisten local` results in:

```
thread 'main' panicked at src/main.rs:20:9:
Unexpected arguments: ["local"]
```

even though it's a valid xwayland option

How can I prevent it from listening on abstract sockets?

Supreeeme:
There's currently no way to do this (without code changes) - the only argument satellite expects and will understand is a display number. But I suppose it would be reasonable to allow passing through all arguments to Xwayland.

bbb651:
We're eventually going to need arguments for xwayland-satellite itself, I can think of two ways arguments could work:
- `xwayland-satellite <args> -- <Xwayland args>` (could also be a single argument value but that feels purely worse, requiring quoting then `shlex`ing it back to separate arguments)
- `xwayland-satellite <mixed args>` where single dash args (except `-h`) are passed as is to Xwayland

Also certain Xwayland args should be explicitly disallowed like `-wm` and `-displayfd`

Supreeeme:
I believe this was actually fixed with #342.

PR #342, opened 2026-01-08
Add command-line options

En-En-Code:
Resolves the needs of #75, #259 and #335 by providing a trimmed-down set of `Xserver` flags which get forwarded to `Xwayland` when passed as additional parameters to `satellite`. Basic validation is done to prevent misuse (such as confirming the `-auth` parameter is a readable file and that `-extension` does not disable an extension mandatory to `satellite`'s functioning).

I also documented the set of supported flags, and their quirks when used specifically with `xwls` in a man page.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -35,6 +35,7 @@ struct ParsedFlags {
     coredump: bool,
     extension_plus: Vec<String>,
     extension_minus: Vec<String>,
+    glamor: Option<&'static str>,
     listen_plus: Vec<String>,
     listen_minus: Vec<String>,
     verbosity: u32,
@@ -48,6 +49,7 @@ impl Default for ParsedFlags {
             coredump: false,
             extension_plus: vec![],
             extension_minus: vec![],
+            glamor: None,
             listen_plus: vec![],
             listen_minus: vec![],
             verbosity: 0,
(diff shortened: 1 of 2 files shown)


src/main.rs
```diff
@@ -20,46 +46,191 @@ impl xwayland_satellite::RunData for RealData {
...
-    let mut i = 2;
-    while i < args.len() {
-        let arg = &args[i];
-        if arg == "-listenfd" {
-            let next = i + 1;
-            if next == args.len() {
-                // Matches the Xwayland error message.
-                panic!("Required argument to -listenfd not specified");
+    // All other options (including the first if it was not a display name) are supported flags
+    while let Some(arg) = args.get(i) {
```
Supreeeme: Could replace this with pure iterator, i.e.
```rust
while let Some(arg) = args.next() {
            match arg.as_str() {
            "-ac" => {
                data.disable_ac = true;
            }
            "-audit" => {
                data.audit_level = args
                    .by_ref()
                    .next()
                    .and_then(|n| n.parse().ok())
                    .expect("argument to -audit not provided or integer");
            }
}
```

xwayland-satellite.man
```diff
@@ -0,0 +1,104 @@
+.TH XWAYLAND\-SATELLITE 1
```
Supreeeme: Did you handwrite this or was this generated with something? I'm not really familiar with manpages but it would be nice to have one.

En-En-Code: I hand-wrote it. `roff` is an old and weird file format, so I kept it super simple. There are certainly man page generators if you wanted to work in an easier-to-use format, but then there is the complexity of adding that generator to the project and making sure the files are synched.

En-En-Code:
Think everything is ready. This can be squashed honestly.

PR #343, opened 2026-01-10
Advertise _NET_WM_MOVERESIZE

YaLTeR:
Makes GTK 4 use compositor-side move and resize.

Disclaimer: idk anything about X11; @bugaevc suggested me that GTK 4 can do compositor-side move/resize, so we checked in gdb, and this was the atom that it was checking the presence of.


Supreeeme:
lgtm

PR #357, opened 2026-01-24
nix refactor: add overlay, remove flake-parts, and other changes

stefanboca:
I noticed the flake didn't have an overlay, so I added one. I also made some other changes while I was at it. Most of the changes are heavily inspired by the structure of Niri's flake.nix.
- add overlay
- add formatter for nix
- remove flake-parts
- use rust-overlay only for the dev shell. end users can set `inputs.rust-overlay.follows = ""` to skip downloading it.
- add rust-analyzer and rust-src extensions in the dev shell
- filter the source in order to reduce rebuilds
- support all Linux systems that the nixpkgs flake exposes
- override VERGEN_GIT_DESCRIBE to include the git rev as well
- remove `clangStdenv` from the dev shell.  I'm not sure what purpose it served. If the intention was faster linking with lld, then it's not needed anymore, now that lld has become the default linker in rust 1.90.0. 

I noticed that `xcb-util-cursor` propagates `libxcb`, so `libxcb` could be removed from `buildInputs`, but decided to keep it in order to be explicit, as it's a direct dependency.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -2,11 +2,11 @@
   "nodes": {
     "nixpkgs": {
       "locked": {
-        "lastModified": 1764522689,
-        "narHash": "sha256-SqUuBFjhl/kpDiVaKLQBoD8TLD+/cTUzzgVFoaHrkqY=",
+        "lastModified": 1769089682,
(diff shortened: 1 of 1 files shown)


Supreeeme:
Don't know anything about Nix, but lgtm.

PR #367, opened 2026-01-31
feat: handle global removals, recalc output scale

### LINKED ISSUE #351: DPI should switch if main screen is HiDPI and last external output is removed, workaround kill xwayland-satellite and restart app
### Operating System

Arch Linux

### Wayland Compositor

Niri

### Describe the issue.

Following scenario. I have a laptop with the resolution 2560x1600, with 2x scaling.

1. I start an xwayland application and the scaling seems good.
2. I attach the external output with a scaling of 1.4 and the dpi adjusts to the external monitor. 
3. On the laptop screen things are now too small, but that's not an issue
4. After detaching the dpi should switch back to the internal monitors 2x scaling
5. To fix the scaling, I have to do the following:
- close the app
- kill xwayland-satellite
- restart app

### Steps to reproduce

_No response_

En-En-Code:
All `GlobalRemove` events sent from the server are now handled by recording them in a new clientside `vec` and passing the identifier returned by `create_global` (now stored by a map in the state) to `disable_global`. `handle_globals` (the top-level function) and `handle_new_globals` (the `InnerServerState` member function) have swapped names to better represent their new purposes.

This enables action to be taken when globals are removed. In this case, the desired action is to forward output removal, so that the scaling calculation does not account for disconnected monitors in its logic.

Resolves #351.

I'm expecting it needs some refactoring, but with a working test (which ended up being the harder part to write) that should be easy enough.

### DIFF
--- a/src/server/clientside.rs
+++ b/src/server/clientside.rs
@@ -206,19 +206,22 @@ impl Dispatch<WlRegistry, GlobalListContents> for MyWorld {
         _: &wayland_client::Connection,
         _: &wayland_client::QueueHandle<Self>,
     ) {
-        if let Event::<WlRegistry>::Global {
-            name,
-            interface,
-            version,
-        } = event
-        {
-            state.new_globals.push(Global {
+        match event {
+            Event::<WlRegistry>::Global {
                 name,
                 interface,
                 version,
-            });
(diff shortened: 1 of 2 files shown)


fishman:
xwayland crashes when disconnecting output now

Issue #247, opened 2025-10-02
[Bug]Non-integer scaling may have slight size deviations in some Windows


DreamMaoMao:
### Operating System

Arch

### Wayland Compositor

sway

### Describe the issue.

When using non-integer scaling, the surface of some applications may be a bit too large, causing misalignment. As seen in the video, when using 1.5 scaling, the sureface in the input method panel exceeds the broder below it. However, this issue does not occur when scaling with integers like 2,3, or 4

### Please confirm you have reproduced this on the latest commit.

- [x] I have reproduced this on the latest commit

DreamMaoMao:
# 1.5 scale

https://github.com/user-attachments/assets/effcef01-f47f-4afa-8970-03541c790a98


# 2 scale

https://github.com/user-attachments/assets/0673130c-2e4e-4e07-a920-6f65137349ca

Supreeeme:
Could you get a log with `RUST_LOG=debug`?

DreamMaoMao:
[test.log](https://github.com/user-attachments/files/22673493/test.log)

here you are.

Supreeeme:
Sorry, to make sure I understand the issue right, the problem is that on the 1.5 scale, the input window is expanding to the left of the cursor, while on the 2 scale, it's under the cursor the whole time?

DreamMaoMao:
No, at 1.5 scale, the entire surface will be slightly stretched and enlarged by approximately 1px.

When zooming in at 2x, you can see that the height of the panel remains unchanged when refreshed, but when zooming in at 1.5x, the height of the panel keeps jumping up and down

Supreeeme:
Oh I see now, the border looks like it's disappearing.

MSCMDD:
Same problem

In some games(eg. Vic3/HOI4), full-screen mode combined with HiDPI (non-integer scale factor) may cause window misalignment, resulting in the mouse being detected as leaving the window when positioned at the screen edges.
It's worth noting that after I executed the command niri msg output eDP-1 scale 1.0, the mouse detection in the game returned to normal.

Command `niri msg windows` Return
```
Window ID 57:
  Title: "Victoria 3"
  App ID: "victoria3"
  Is floating: no
  PID: 984
  Workspace ID: 1
  Layout:
    Tile size: 1706.67 x 1066.67
    Scrolling position: column 4, tile 1
    Window size: 1706 x 1066
    Window offset in tile: 0.67 x 0.67
```

[...]

PR #374, opened 2026-02-11
fix: Round up the width under HIDPI

MSCMDD:
Fixed issue where the fullscreen width was incorrectly rounded down under HiDPI scaling

https://github.com/niri-wm/niri/issues/3417
https://github.com/Supreeeme/xwayland-satellite/issues/247#issuecomment-3869547851


MSCMDD:
@Supreeeme emmmm fixed...

PR #378, opened 2026-02-17
fix: align global output offset with screen edges

### LINKED ISSUE #340: Tablet: pen, pointer offset (in Krita)
### Operating System

NixOS-25.11

### Wayland Compositor

Niri

### Describe the issue.

When using Krita to draw or select tools, there is a horizontal offset of where niri would map the cursor and where Krita detects it.\
It could be a me issue, due to Krita working correctly a few days ago and me not finding any change in version of the software (I think is) involved.

```txt
#xprop
_VARIABLE_REFRESH(CARDINAL) = 1
WM_STATE(WM_STATE):
		window state: Normal
		icon window: 0x0
_NET_WM_USER_TIME(CARDINAL) = 788855
WM_WINDOW_ROLE(STRING) = "MainWindow#1"
_NET_WM_ICON_NAME(UTF8_STRING) =
_NET_WM_ICON(CARDINAL) = 	Icon (16 x 16): # ... (the icons)
XdndAware(ATOM) = BITMAP
WM_NAME(STRING) =
_NET_WM_NAME(UTF8_STRING) = "Krita"
_MOTIF_WM_HINTS(_MOTIF_WM_HINTS) = 0x3, 0x3e, 0x7e, 0x0, 0x0
_NET_WM_WINDOW_TYPE(ATOM) = _NET_WM_WINDOW_TYPE_NORMAL
_XEMBED_INFO(_XEMBED_INFO) = 0x0, 0x1
WM_CLIENT_LEADER(WINDOW): window id # 0x40000e
WM_HINTS(WM_HINTS):
		window id # of group leader: 0x40000e
WM_CLIENT_MACHINE(STRING) = "legionOfNix"
_NET_WM_PID(CARDINAL) = 12420
_NET_WM_SYNC_REQUEST_COUNTER(CARDINAL) = 4194338
WM_CLASS(STRING) = "krita", "krita"
WM_PROTOCOLS(ATOM): protocols  WM_DELETE_WINDOW, WM_TAKE_FOCUS, _NET_WM_PING, _NET_WM_SYNC_REQUEST
		user specified location: 3200, 0
		user specified size: 1828 by 1070
		program specified minimum size: 492 by 480
		window gravity: Static
```

```txt
#niri msg output
Window ID 25:
  Title: "Krita"
  App ID: "krita"
  Is floating: no
  PID: 10593
  Workspace ID: 2
  Layout:
    Tile size: 1828 x 1070
    Scrolling position: column 3, tile 1
    Window size: 1828 x 1070
    Window offset in tile: 0 x 0
```

### Steps to reproduce

Open Krita and start drawing towards the cavas sides.

En-En-Code:
Previously, global output offset only moved outputs if they were in negative coordinate space. Doing so guaranteed an output on the top-most and left-most border of the screen. This commit offsets outputs to achieve the same goal on outputs in positive coordinate space, which matches the output placement of `xrandr`, a placement assumption made by Krita's tablet tool logic.

The new definition of global output offset can be thought of as drawing the smallest possible rectangle with contains every monitor, and moving that bounding box to (0, 0).

Resolves #340.

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -34,13 +34,16 @@ The code for the X11 portion of satellite lives in `src/xstate`. Satellite must 
 as any other standard X11 window manager. This includes:
 
 - Setting SubstructureRedirect and SubstructureNotify on the root window, to get notifications for when new windows are being created
-- Follwing (most of) the [ICCCM](https://www.x.org/releases/X11R7.6/doc/xorg-docs/specs/ICCCM/icccm.html) and [EWMH](https://specifications.freedesktop.org/wm-spec/latest/) specs
+- Following (most of) the [ICCCM](https://www.x.org/releases/X11R7.6/doc/xorg-docs/specs/ICCCM/icccm.html) and [EWMH](https://specifications.freedesktop.org/wm-spec/latest/) specs
 
 In addition, satellite must do some other things that a normal X11 window manager wouldn't - but a compositor integrating
 Xwayland would - such as synchronize X11 and Wayland selections. This is explained further in the Wayland server section.
 
 The way that satellite manages windows from the X11 point of view is as follows:
 
+- All monitors maintain their relative positions to one another. Their absolute position is such that
+  the top-most monitor's top edge is on the X-axis and the left-most monitor's left edge is on the Y-axis.
+    - All monitors are on non-negative coordinates with no gaps between the screen and any monitor, matching what `xrandr` does.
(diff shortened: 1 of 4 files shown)


Supreeeme:
This is strange to me - I would expect Xwayland itself to handle this assumption. Then again, I would also expect Xwayland to handle negative output coordinates but it doesn't do that either. I also find it strange that Krita is even reliant on this behavior, but if it fixes a bug it fixes a bug. I don't have a real reason to object to this PR, but I wonder if it would be worth bringing up this behavior as an Xwayland bug.

Issue #331, opened 2025-12-24
Show client-drawn popup styles.


busyoGG:
**Is your feature request related to a problem? Please describe.**

All popups cannot show client-drawn styles.

**Describe the solution you'd like**

All popups can show client-drawn styles.

**Describe alternatives you've considered**

**Additional context**

### current

<img width="506" height="517" alt="Image" src="https://github.com/user-attachments/assets/ce3bd7ce-a331-4093-829c-50b7926fc394" />

### normal one

<img width="514" height="540" alt="Image" src="https://github.com/user-attachments/assets/98759c83-64fe-46a8-8a22-9e164e0edb73" />

YaLTeR:
I suspect this is due to xwl-s not advertising alpha compositing support in some way (i have no idea what i'm talking about here)

busyoGG:
Fixed by #381

PR #381, opened 2026-02-20
fix: take ownership of `_NET_WM_CM_S0`

### LINKED ISSUE #376: Some Avalonia UI apps do not have alpha transparency, leading to white backgrounds on certain UI controls
### Operating System

CachyOS

### Wayland Compositor

Niri

### Describe the issue.

When loading an app based on [Avalonia UI](https://avaloniaui.net/), namely with the Semi.Avalonia theme which uses shadows, some elements do not have transparency when they should, leading to white backgrounds where there shouldn't be (e.g. menus, combo boxes).

Image

Meanwhile, this does not occur on KDE's implementation of XWayland.

Image

I haven't been able to see if there were any parameters I could pass to xwayland-satellite to fix this.

Related issue: https://github.com/irihitech/Semi.Avalonia/issues/682

### Steps to reproduce

1. Load an application that uses Semi.Avalonia + menus/combo box
2. Observe the lack of transparent background on top menus or combo boxes.

**Expected behaviour:**
Background should be transparent.

En-En-Code:
EWMH section 8.1 requires that compositing window managers take ownership of `_NET_WM_CM_Sn` for each screen `n` they manage (in the case of `satellite`, this is just screen 0).

The Avalonia UI toolkit needs this atom to be owned in order to provide transparent window support. Resolves #376.

### DIFF
--- a/src/xstate/mod.rs
+++ b/src/xstate/mod.rs
@@ -333,6 +333,14 @@ impl XState {
                 time: x::CURRENT_TIME,
             })
             .unwrap();
+
+        self.connection
+            .send_and_check_request(&x::SetSelectionOwner {
+                owner: self.wm_window,
+                selection: self.atoms.net_wm_cm_s0,
+                time: x::CURRENT_TIME,
+            })
+            .unwrap();
     }
 
     pub fn handle_events(&mut self, server_state: &mut super::RealServerState) {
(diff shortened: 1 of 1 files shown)


En-En-Code:
Related to window transparency, I fast-forwarded the old alpha-modifier PR. Was not hard (annoying, perhaps) outside of needing to reimplement `set_wm_opacity` since it was written back in the `HopSlotMap` days. Tested it on Sway by using `xprop` to set `_NET_WM_WINDOW_OPACITY` so it does work, even if I still do not know of a real-world use of it. If you want it, I can put it up.

Supreeeme:
So transparent windows just start working with Avalonia with this change and without `_NET_WM_WINDOW_OPACITY` support?

En-En-Code:
At least the problem window in the issue does. I checked its `xprop` and it does not even set `_NET_WM_WINDOW_OPACITY`, so I am not sure what it does. Probably some code specifically for a 0-opacity window is in place.

busyoGG:
This also fixes  #331

PR #390, opened 2026-02-28
Refactor window heuristics

En-En-Code:
Finally got back around to cleaning this mess up. Did I lot more than I thought I would lol.

Also, the `SPLASH` handling is inspired by a real-world X application (NASPA Zyzzyva) having a splash screen without the size bounds and being tiled like a regular toplevel in Niri/Sway, which looked bad.


src/xstate/mod.rs
```diff
@@ -623,44 +623,61 @@ impl XState {
...
         }
-
         if let Some(name) = title {
             server_state.set_win_title(window, name);
         }
+
+        let class = self.get_wm_class(window);
         if let Some(class) = class.resolve()? {
             server_state.set_win_class(window, class);
         }
```
Supreeeme: If the properties are being instantly resolved like this, then there's probably no point to the PropertyCookieWrapper anymore.

src/xstate/tests.rs
```diff
@@ -0,0 +1,279 @@
...
+    // Originally, `skip_taskbar` was used to check the pop-up status of the dialog, but since then,
+    // override_redirect with no NET window type (fallback to NORMAL) is the used heuristic
+    #[test]
+    fn reaper_dialog() {
+        let win_types = WindowTypes::new();
+        let win = WindowRoleHeuristics {
+            override_redirect: true,
+            wm_hints: Some(WmHints::from(
+                [0x067_u32, 1, 1, 0x4000128, 0, 0, 0, 0x400012e, 0x4000001].as_slice(),
+            )),
```
Supreeeme: I would rather these tests did not use magic numbers in the properties, it makes it hard to process what they're supposed to be testing for.

src/xstate/mod.rs
```diff
@@ -1241,6 +1102,133 @@ mod motif {
...
+            if hints.functions.is_some_and(|f| f.is_empty()) {
+                return WindowRole::Popup;
+            }
+        }
+        if let Some(hints) = self.wm_normal_hints {
+            forced_size = hints
+                .min_size
+                .zip(hints.max_size)
+                .is_some_and(|(min, max)| min == max);
+        }
```
Supreeeme:
```suggestion
        let forced_size = self.wm_normal_hints.is_some_and(|hints| {
           hints
                .min_size
                .zip(hints.max_size)
                .is_some_and(|(min, max)| min == max)
        });
```

src/xstate/tests.rs
En-En-Code: Yeah, all of the hex values except the first (the bitflags) are X windows, which creates some noise. My goal was to have the windows exactly as their `xprop` represented them, for effective regresion testing. Some properties we do not handle yet, so the `From` constructor was the way to future-proof for that.

En-En-Code: Had a 3 AM realization that the builder pattern is probably the right choice for this. I was concerned it would feel like "Abstraction for the Abstraction Gods," but it is much more readable.

En-En-Code:
Small change to avoid a regression with Krita's dialog windows. With set `_MOTIF_WM_HINTS`, they should be top-levels, since pop-ups cannot be decorated.
```
WM_STATE(WM_STATE):
		window state: Normal
		icon window: 0x0
_NET_WM_USER_TIME(CARDINAL) = 90356019
WM_TRANSIENT_FOR(WINDOW): window id # 0x400020
_NET_WM_ICON_NAME(UTF8_STRING) =
_NET_WM_ICON(CARDINAL) = 	-- snip --
XdndAware(ATOM) = BITMAP
WM_NAME(STRING) = "Brush Editor"
_NET_WM_NAME(UTF8_STRING) = "Brush Editor — Krita"
_MOTIF_WM_HINTS(_MOTIF_WM_HINTS) = 0x3, 0x26, 0x1e, 0x0, 0x0
_NET_WM_WINDOW_TYPE(ATOM) = _NET_WM_WINDOW_TYPE_DIALOG, _NET_WM_WINDOW_TYPE_NORMAL
_XEMBED_INFO(_XEMBED_INFO) = 0x0, 0x1
WM_CLIENT_LEADER(WINDOW): window id # 0x400009
WM_HINTS(WM_HINTS):
		window id # of group leader: 0x400009
WM_CLIENT_MACHINE(STRING) = "EnEnComputer"
_NET_WM_PID(CARDINAL) = 4
_NET_WM_SYNC_REQUEST_COUNTER(CARDINAL) = 4194781
_GTK_APPLICATION_ID(UTF8_STRING) = "org.kde.krita"
_KDE_NET_WM_DESKTOP_FILE(UTF8_STRING) = "org.kde.krita"
WM_CLASS(STRING) = "krita", "krita"
WM_PROTOCOLS(ATOM): protocols  WM_DELETE_WINDOW, WM_TAKE_FOCUS, _NET_WM_PING, _NET_WM_SYNC_REQUEST
		user specified location: 638, 355
		user specified size: 1382 by 709
		program specified minimum size: 827 by 709
		window gravity: Static
```
In addition, there are similar windows which also set `_NET_WM_STATE` to include `_NET_WM_STATE_MODAL`, such as the "Save As" menu. These windows are different in that they force focus onto themselves instead of letting the user interact with the main top-level. `MODAL` has in the past been considered a pop-up weed-out (though insufficient in this particular case).

src/xstate/tests.rs
```diff
@@ -20,10 +20,158 @@ impl WindowTypes {
...
+    /// If called after `program_pos`, both positions will be set to the value passed to `user_pos`.
+    fn user_pos(mut self, x: u32, y: u32) -> Self {
+        self.fields[0] |= 0x001;
+        self.fields[1] = x;
+        self.fields[2] = y;
+        self
+    }
+    /// If called after `program_size`, both sizes will be set to the value passed to `user_size`.
+    fn user_size(mut self, w: u32, h: u32) -> Self {
+        self.fields[0] |= 0x002;
```
Supreeeme: I like these builders, but could we reuse the flags instead of these numbers?

src/xstate/tests.rs
```diff
@@ -20,10 +20,158 @@ impl WindowTypes {
...
+    }
+    /// Each tuple's first number is its numerator and second is its denominator.
+    fn _aspect_ratios(mut self, min: (u32, u32), max: (u32, u32)) -> Self {
+        self.fields[0] |= 0x080;
+        self.fields[11] = min.0;
+        self.fields[12] = min.1;
+        self.fields[13] = max.0;
+        self.fields[14] = max.1;
+        self
+    }
```
Supreeeme, line 91: I assume this is unused, does it need to exist?

En-En-Code: It's here for future-proofing. If writing a test for an X window which sets this field of its `WM_NORMAL_HINTS`, it's nice to already have.

src/xstate/tests.rs
En-En-Code: Ultimately, the flags have to be assigned to numbers somewhere. If I add all the missing flags to the `bitflags!`, then it should be fine.

En-En-Code:
At a strange place right now with this. In a conversation (in #417), I realized the `WM_HINTS` pop-up heuristic has caused more problems than it has fixed. It's within the scope of this PR to revert that, although leaving Yabridge with regressions is not ideal. We could also hard-code an exception for Yabridge (see #315) in this. Either way, I probably want to rebase this (being broken into multiple commits is useful, but the additional ones are mostly noise).l

Supreeeme:
I can live with removing the WM_HINTS heuristics and hardcoding a yabridge exception.

Issue #278, opened 2025-11-13
Can't type in text fields in Unity editor popups


justDeeevin:
### Operating System

NixOS 25.11

### Wayland Compositor

Niri

### Describe the issue.

When working in the Unity editor, certain popups with text fields cannot be typed into. It seems like the window beneath takes focus back when I try to type. So far this has prevented me from typing in a URL field when trying to install a package from a git remote and typing in a name for a new entity tag.

### Steps to reproduce

In the Unity editor, create a new tag, and try to type in a name for it.

### Please confirm you have reproduced this on the latest commit.

- [x] I have reproduced this on the latest commit

Supreeeme:
Can you provide a log? Run satellite with `RUST_LOG=debug` set and then reproduce the issue.

justDeeevin:
Hi sorry for not getting back to you for so long

How can I set that environment variable for xwayland-satellite if it's spawned by niri?

ToxicMushroom:
Niri has an environment block
```kdl
environment { 
    RUST_LOG "debug"
}
```
But this will most likely cause log spewing by all your rust apps, I'd recommend launching xwl-s manually for this.
You can disable the niri integration as such:
```kdl
xwayland-satellite {
    off
}
environment {
    DISPLAY ":0"
}
```
And to run it something like:
`RUST_LOG=debug &> xwayland_satellite.log`

Andicraft:
I'm having the same issue, with Sway as my WM. Attaching a log - I also can't drag and drop tabs or other things like in #286 and I attempted to do so in the logged session as well.

[Xwayland-Satellite Unity Log.txt](https://github.com/user-attachments/files/24016925/Xwayland-Satellite.Unity.Log.txt)

ural89:
I have the same problem. I cannot search and add component, so i cannot daily drive niri

revanj:
Related to https://github.com/YaLTeR/niri/issues/3244

CheesecakeCG:
I did a bisect, and it seems that e8079bc072c5864b1785250b201c0a77c08d6355 caused this issue.

Supreeeme:
Can someone provide a screenshot of the popups, and the output of running `xprop` on one?

Fireye04:
> Can someone provide a screenshot of the popups, and the output of running `xprop` on one?

[...]

PR #397, opened 2026-03-12
Fix: unity editor add component pop up

WMsans:
Solves #278

### DIFF
--- a/src/server/event.rs
+++ b/src/server/event.rs
@@ -620,17 +620,10 @@ impl Event for client::wl_pointer::Event {
                 let mut query = surface_entity.and_then(|e| {
                     state
                         .world
-                        .query_one::<(
-                            &WlSurface,
-                            &SurfaceRole,
-                            &SurfaceScaleFactor,
-                            &x::Window,
-                            &WindowData,
-                        )>(e)
+                        .query_one::<(&WlSurface, &SurfaceRole, &SurfaceScaleFactor, &x::Window)>(e)
                         .ok()
                 });
-                let Some((surface, role, scale, window, window_data)) =
-                    query.as_mut().and_then(|q| q.get())
+                let Some((surface, role, scale, window)) = query.as_mut().and_then(|q| q.get())
                 else {
                     if let Some(&DecorationMarker { parent }) = surface.data() {
                         drop(query);
@@ -649,17 +642,12 @@ impl Event for client::wl_pointer::Event {
                 cmd.insert(target, (*scale,));
 
                 let surface_is_popup = matches!(role, SurfaceRole::Popup(_));
-                let popup_wants_focus = surface_is_popup && window_data.attrs.acquire_input_via_wm;
                 let mut do_enter = || {
                     debug!("pointer entering {} ({serial} {})", surface.id(), scale.0);
                     server.enter(serial, surface, surface_x * scale.0, surface_y * scale.0);
                     connection.raise_to_top(*window);
                     if !surface_is_popup {
                         state.last_hovered = Some(*window);
-                    }
-                    if popup_wants_focus {
-                        debug!("focusing popup {:?} (acquire_input_via_wm)", window);
-                        connection.focus_window(*window, None);
                     }
                     cmd.insert_one(target, CurrentSurface::Xwayland(surface_entity.unwrap()));
                 };
@@ -702,31 +690,6 @@ impl Event for client::wl_pointer::Event {
                 {
                     decoration::handle_pointer_leave(state, parent);
                     return;
-                }
-
-                // If we're leaving a popup that acquired input focus, restore
-                // focus to the last focused toplevel.
-                if let Some(entity) = surface.data().copied() {
-                    let should_restore_focus = state
-                        .world
-                        .query_one::<(&SurfaceRole, &WindowData)>(entity)
-                        .ok()
-                        .and_then(|mut q| {
-                            q.get().map(|(role, wd)| {
-                                matches!(role, SurfaceRole::Popup(_))
-                                    && wd.attrs.acquire_input_via_wm
-                            })
-                        })
-                        .unwrap_or(false);
-                    if should_restore_focus {
-                        if let Some(toplevel) = state.inner.last_focused_toplevel {
-                            debug!(
-                                "restoring focus to toplevel {:?} after leaving popup",
-                                toplevel
-                            );
-                            state.connection.focus_window(toplevel, None);
-                        }
-                    }
                 }
 
                 if let Some(surface) = surface
@@ -802,12 +765,14 @@ impl Event for client::wl_pointer::Event {
                     pub const LEFT: u32 = 0x110;
                 }
 
+                let mut popup_to_focus = None;
                 if button_state == WEnum::Value(client::wl_pointer::ButtonState::Pressed)
                     && button == button_codes::LEFT
                 {
                     match current_surface {
                         CurrentSurface::Xwayland(entity) => {
                             cmd.insert(*entity, (LastClickSerial(seat.clone(), serial),));
+                            popup_to_focus = Some(*entity);
                         }
                         CurrentSurface::Decoration(parent) => {
                             let seat = seat.clone();
@@ -821,6 +786,36 @@ impl Event for client::wl_pointer::Event {
 
                 server.button(serial, time, button, convert_wenum(button_state));
                 drop(query);
+
+                // Focus popups that want input on click, so that X11 clients
+                // with text fields (e.g. Unity's "Add Component" search bar)
+                // can receive keyboard input.
+                if let Some(entity) = popup_to_focus {
+                    let should_focus = state
+                        .world
+                        .query_one::<(&SurfaceRole, &WindowData, &x::Window)>(entity)
+                        .ok()
+                        .and_then(|mut q| {
+                            q.get().map(|(role, wd, window)| {
+                                if matches!(role, SurfaceRole::Popup(_))
+                                    && wd.attrs.acquire_input_via_wm
+                                {
+                                    Some(*window)
+                                } else {
+                                    None
+                                }
+                            })
+                        })
+                        .flatten();
+                    if let Some(window) = should_focus {
+                        debug!(
+                            "focusing popup {:?} on click (acquire_input_via_wm)",
+                            window
+                        );
+                        state.connection.focus_window(window, None);
+                    }
+                }
+
                 cmd.run_on(&mut state.world);
             }
             _ => {
--- a/src/server/event.rs
+++ b/src/server/event.rs
@@ -620,10 +620,17 @@ impl Event for client::wl_pointer::Event {
                 let mut query = surface_entity.and_then(|e| {
                     state
                         .world
-                        .query_one::<(&WlSurface, &SurfaceRole, &SurfaceScaleFactor, &x::Window)>(e)
+                        .query_one::<(
+                            &WlSurface,
+                            &SurfaceRole,
+                            &SurfaceScaleFactor,
+                            &x::Window,
+                            &WindowData,
+                        )>(e)
                         .ok()
                 });
-                let Some((surface, role, scale, window)) = query.as_mut().and_then(|q| q.get())
+                let Some((surface, role, scale, window, window_data)) =
+                    query.as_mut().and_then(|q| q.get())
                 else {
                     if let Some(&DecorationMarker { parent }) = surface.data() {
                         drop(query);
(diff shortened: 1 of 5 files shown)


src/server/event.rs
```diff
@@ -783,6 +786,36 @@ impl Event for client::wl_pointer::Event {
...
+                                if matches!(role, SurfaceRole::Popup(_))
+                                    && wd.attrs.acquire_input_via_wm
+                                {
+                                    Some(*window)
+                                } else {
+                                    None
+                                }
+                            })
+                        })
+                        .flatten();
```
En-En-Code: Use `query_one_mut` instead to simpilfy this call chain.

src/server/mod.rs
```diff
@@ -944,13 +945,15 @@ impl<S: X11Selection + 'static> InnerServerState<S> {
         }
     }
 
-    pub fn set_win_hints(&mut self, window: x::Window, hints: WmHints) {
+    pub fn set_win_hints(&mut self, window: x::Window, hints: &WmHints) {
```
En-En-Code: Keep the original signature and make `WmHints` `Clone`/`Copy` instead.

src/server/tests.rs
En-En-Code, line 1: Make sure this files (and all others) are formatted with Rust's formatting standard.

En-En-Code (changes requested):
After mulling over [the input focus specification](https://tronche.com/gui/x/icccm/sec-4.html#s-4.1.7) a bit, the current input focus handling feels off to me but I can't put my finger on why. This change is sensible enough that I think requesting a refactor to the system for what are, as of now, only theoretical issues is unwarranted.

If you are up for it, writing a test demonstrating the changes would be valuable.

WMsans:
Thank you for your feedback! I'll work on it today. I don't have much experience with tests but I'll try my best =)

> The current input focus handling feels off to me but I can't put my finger on why. 

The focus logic is that the popups are focused when the user left-clicks on them. In addition, focus would not be restored  to last_focused_toplevel after the pointer left an acquire_input_via_wm popup. I can see some issues that may caused by this, like some popups that need hover interaction only instead of clicking. 

For the ICCCM, I didn't use any of the input model. It is designed for Unity Editor's specific case. I'll try to write some test to demonstrate this, or come up with a work-around that matches the standards. Please let me know your concerns!

testwl/src/lib.rs
```diff
@@ -851,6 +851,22 @@ impl Server {
         self.display.flush_clients().unwrap();
     }
 
+    pub fn click_pointer(&mut self, _surface: SurfaceId, button: u32) {
```
En-En-Code: `#[track_caller]`, also, remove the unneeded parameter.

testwl/src/lib.rs
```diff
@@ -851,6 +851,22 @@ impl Server {
...
+        let pointer = &self
+            .state
+            .pointer
+            .as_ref()
+            .expect("No pointer created")
+            .pointer;
+        let time = self.state.begin.elapsed().as_millis() as u32;
+        let serial = self.state.configure_serial;
+        self.state.configure_serial += 1;
```
En-En-Code:
```diff
-- self.state.configure_serial += 1;
```
Should not need this. It was originally intended to count window reconfigurations, which this is not.

src/server/tests.rs
```diff
@@ -1607,6 +1607,82 @@ fn override_redirect_choose_hover_window() {
     assert_eq!(&popup_data.popup().parent, win1_xdg);
 }
 
+/// Clicking on a popup that has `acquire_input_via_wm` set should focus
+/// that popup window, so that X11 clients with input fields inside popups
+/// (e.g. Unity Editor's "Add Component" search bar) receive keyboard input.
+#[test]
+fn click_focuses_popup_with_acquire_input_via_wm() {
```
En-En-Code: Minor: I would combine these tests (first without, then with `acquire_input`), then give it a more succinct name.

En-En-Code (changes requested):

En-En-Code:
> For the ICCCM, I didn't use any of the input model.

Yeah, the ICCCM is an obtuse document. I do not expect newcomers to understand what it wants, and I still don't sometimes either. The scope of this PR is to resolve the Unity search box issue in a "minimal" way.

When I said the focus handling is off, that is because I think what it wants from the window manager, at least for "Locally Active" windows, is for the manager to send a `ClientMessage` event, then the client to `SetInputFocus` with `revert_to` set to `Parent` so the focus returns to the toplevel when the popup is unmapped, without the window manager needing to orchestrate that. That's called additional complexity and we'll only deal with it if the need arises.

src/server/event.rs
```diff
@@ -783,6 +786,28 @@ impl Event for client::wl_pointer::Event {
 
                 server.button(serial, time, button, convert_wenum(button_state));
                 drop(query);
+
+                // Focus popups that want input on click, so that X11 clients
+                // with text fields (e.g. Unity's "Add Component" search bar)
+                // can receive keyboard input.
```
Supreeeme: Focusing should not be done on click - not all compositors do focusing only when clicked (i.e. any compositor with a focus-follows-mouse mode), and the focus of interest here is not mouse focus but "input focus", i.e. keyboard focus. I haven't dug into this at all, but it seems that all windows should already be focused on keyboard focus, so the question to address would be why isn't that the case with the Unity popup windows.

testwl/src/lib.rs
```diff
@@ -851,6 +851,22 @@ impl Server {
         self.display.flush_clients().unwrap();
     }
 
+    #[track_caller]
+    pub fn click_pointer(&mut self, button: u32) {
```
Supreeeme: Would greatly prefer an enum for the button, the numbers are pretty much meaningless when just scanning through.

src/server/mod.rs
```diff
@@ -100,6 +100,7 @@ where
 #[derive(Default, Debug)]
 struct WindowAttributes {
     is_popup: bool,
+    acquire_input_via_wm: bool,
```
Supreeeme, line 103:
```suggestion
    allow_focus_when_popup: bool,
```

WMsans:
> so the question to address would be why isn't that the case with the Unity popup windows.

After some research, in my opinion, the issue is that these are override-redirect X11 windows with WM_HINTS input: True, which get created as xdg_popup surfaces on the Wayland side. As a result, compositors don't send wl_keyboard.enter to xdg_popup surfaces as the keyboard focus stays on the parent. 

Thus, maybe setting focus when WM_HINTS input: True is set, instead of on mouse clicking, would solve the issue?

src/server/mod.rs
```diff
@@ -735,6 +738,13 @@ impl<C: XConnection> ServerState<C> {
                 self.connection.focus_window(x::WINDOW_NONE, None);
             }
             self.unfocus = false;
+
+            // Focus popups that set WM_HINTS input=True at map time, matching
+            // what a real X11 WM would do (call SetInputFocus on map).
+            if let Some(popup_win) = self.popup_needs_focus.take() {
+                debug!("focusing popup {popup_win:?} at map time (allow_focus_when_popup)");
+                self.connection.focus_window(popup_win, None);
+            }
```
Supreeeme: It would make more sense to reuse the `to_focus` machinery above, maybe adding an `is_popup` field to `FocusData`.

src/server/mod.rs
```diff
@@ -1406,6 +1428,16 @@ impl<S: X11Selection + 'static> InnerServerState<S> {
...
 
+        // If an X11 popup set WM_HINTS input=True (allow_focus_when_popup),
+        // it expects the window manager to give it keyboard focus via
+        // SetInputFocus. Record this so the caller (which has access to the
+        // XConnection) can call focus_window. We do not use xdg_popup.grab()
+        // because grab() has dismissal semantics (PopupDone → unmap) that
+        // X11 clients don't expect.
+        if focus_popup && popup_for.is_some() {
+            self.popup_needs_focus = Some(window);
+        }
```
Supreeeme: I think this would be better placed in the `configure` event of the popup (in src/server/event.rs)

src/xstate/mod.rs
```diff
@@ -1145,10 +1146,10 @@ impl From<&[u32]> for WmNormalHints {
     }
 }
 
-#[derive(Default, Debug, PartialEq, Eq)]
+#[derive(Copy, Clone, Default, Debug, PartialEq, Eq)]
 pub struct WmHints {
     pub window_group: Option<x::Window>,
-    pub acquire_input_via_wm: bool,
+    pub allow_focus_when_popup: bool,
```
Supreeeme: I think this one should stay as `acquire_input_via_wm`, just so it's obvious that it's matching the WmHints fields.

Supreeeme (commented):
Sorry for the delay. Looking good.

src/server/tests.rs
```diff
@@ -1613,6 +1613,67 @@ fn override_redirect_choose_hover_window() {
...
+fn popup_no_focus_without_input_hint() {
+    let (mut f, comp) = TestFixture::new_with_compositor();
+    TestObject::<WlPointer>::from_request(&comp.seat.obj, wl_seat::Request::GetPointer {});
+
+    let win_toplevel = Window::new(1);
+    let (_, toplevel_id) = f.create_toplevel(&comp, win_toplevel);
+
+    // A popup without acquire_input_via_wm should not receive focus.
+    let win_popup = Window::new(2);
+    let (_, _popup_id) = f.create_popup(
```
En-En-Code: This test does not need the pointer object anymore. Similarly, no need the for the let binding on the popup will not be used.

Could also take the `focused_window` check, move it to `popup_flow_simple`, and remove this test, but unit tests are cheap and fast, so I do not mind small, narrowly-scoped tests.

src/server/mod.rs
```diff
@@ -726,11 +728,17 @@ impl<C: XConnection> ServerState<C> {
...
                 window,
                 output_name,
+                is_popup,
             }) = self.to_focus.take()
             {
-                debug!("focusing window {window:?}");
+                debug!(
+                    "focusing {} {window:?}",
+                    if is_popup { "popup" } else { "window" }
+                );
```
En-En-Code, line 737: Since you are distinguishing it from a popup, call the else case "toplevel"

En-En-Code (changes requested):
Cool, good design progress was made here. Thinking about the Unity editor case in particular, the popup immediately getting input focus means the search box could be typed into without even been clicked (or maybe it cannot, if the search box is its own window or some other strange reason).

Just a couple small things from me.

src/xstate/mod.rs
En-En-Code: I remember originally suggesting this name and not liking how clunky it sounds, but it does do a good job of describing what it is supposed to represent according to the ICCCM (unlike the ICCCM itself, which calls it "input field" or "input hint"). All ears to a better name though.

musjj:
Could we have a new minor release for this PR? There has been quite a few fixes since the last one.

Jurgencotting:
## Regression: this fixes Add Component focus, but breaks Project window rename (and can crash the Editor)

**Environment**
- OS: CachyOS (Arch-based)
- Compositor: Niri (built-in xwayland-satellite integration, spawned with `-listenfd`)
- xwayland-satellite: built from this branch at commit `928799b`
- Unity: Unity 6 (will confirm exact point version if needed)

**What works**
Confirmed the issue this PR targets is fixed: the search field in the "Add Component" popup now receives keyboard input correctly.

**What broke**
Renaming any asset in the Project window stopped working. Neither F2 nor the right-click "Rename" menu option enters edit mode — the label doesn't even become editable. No errors appear in the Console when this happens.

Clicking at specific positions occasionally does trigger edit mode successfully, which points to a focus race rather than a hard failure. After several repeated attempts, the Editor crashed outright (Unity's own crash reporter opened afterward).

**Confirmed via A/B test**
Reverting to the released `xwayland-satellite` 0.8.1-2 package (no other changes) immediately restores normal rename behavior, and reintroduces the original Add Component bug. Rebuilding this branch reproduces the rename regression again. Repeated this swap twice with consistent results.

**Suspected cause**
This PR removes the `has_take_focus` check from the first-configure handling in `popup_event`, so any `xdg_popup` with `WM_HINTS.input = True` now gets keyboard focus forced onto it regardless of whether it also advertises `WM_TAKE_FOCUS`. The previous behavior excluded popups advertising `WM_TAKE_FOCUS` from forced focus (there was a test for this, `popup_no_focus_input_hint_wm_take_focus`, which this PR also removes).

My guess is that Unity's hover tooltips (e.g., the small tooltip showing a file's name when hovering over it in the Project window) are built from the same generic override-redirect popup implementation as the Add Component dropdown, and likely share the same `WM_HINTS`/`WM_TAKE_FOCUS` combination despite not being interactive. If so, every time a tooltip appears while hovering over an asset — which happens right before/during the click used to trigger rename — it would now steal keyboard focus from the Project window, breaking the click sequence Unity uses to enter rename mode. The rapid focus churn from tooltips appearing/disappearing during repeated clicks could plausibly be what crashes the Editor.

I can't inspect Unity's internal popup implementation to confirm this directly, but happy to grab an `xprop` dump of an open tooltip vs. the Add Component popup, or debug logs from a reproduction, if that helps narrow it down.

Thanks for the original fix — Add Component was a real blocker for using Unity under Niri. Flagging this so the rename regression doesn't surprise others testing this branch.

PR #399, opened 2026-03-14
fix: cast ExitStatus to correct type, remove Boxes

En-En-Code:
Absolutely no clue why Past Me thought `usize` was the correct type to convert `ExitStatus` from. Rust docs make it really clear it is a `c_int` (i32) on Unix.

Past Me also decided both leaking memory with `Box::into_raw` then unsoundly calling `Box::from_raw` on a pointer to a stack-allocated array was smart. Dunno why. Loser did not even leave a safety comment smh.

Did I actually author this exactly a year ago? Proves how much I'm still improving, on the plus side.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -132,7 +132,7 @@ pub fn main(mut data: impl RunData) -> Option<()> {
     ];
 
     fn xwayland_exit_code(rx: &mut UnixStream) -> ExitStatus {
-        let mut data = [0; (i32::BITS / 8) as usize];
+        let mut data = [0; std::mem::size_of::<i32>()];
         rx.read_exact(&mut data).unwrap();
         ExitStatus::from_raw(i32::from_ne_bytes(data))
     }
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -121,21 +121,20 @@ pub fn main(mut data: impl RunData) -> Option<()> {
...
 
     let mut ready_fds = [
         PollFd::new(&socket, PollFlags::IN),
         PollFd::new(&finish_rx, PollFlags::IN),
     ];
 
-    fn xwayland_exit_code(rx: &mut UnixStream) -> Box<ExitStatus> {
-        let mut data = [0; (usize::BITS / 8) as usize];
+    fn xwayland_exit_code(rx: &mut UnixStream) -> ExitStatus {
+        let mut data = [0; (i32::BITS / 8) as usize];
```
Supreeeme:
```suggestion
        let mut data = [0; std::mem::size_of<i32>()];
```

PR #403, opened 2026-03-20
server: track global scale to initialize new surfaces correctly

awsms:
Maintain the last known global scale in InnerServerState and use it to initialize the scale factor for new surfaces.
This fixes a race condition where a surface could have its initial size hints calculated with a default 1.0 scale before it officially entered an output, preventing feedback loops that caused windows to double in size on HiDPI displays.

Issue illustrated:

https://github.com/user-attachments/assets/fe15c630-65a0-45c5-b4ae-50a5691a57a7

https://github.com/user-attachments/assets/49906201-a90e-41b1-b3bf-89532c41f81f

### DIFF
--- a/tests/integration.rs
+++ b/tests/integration.rs
@@ -223,6 +223,17 @@ impl Fixture {
         window: x::Window,
         surface: testwl::SurfaceId,
     ) {
+        self.configure_and_verify_new_toplevel_with_size(connection, window, surface, 100, 100);
+    }
+
+    fn configure_and_verify_new_toplevel_with_size(
+        &mut self,
+        connection: &mut Connection,
+        window: x::Window,
+        surface: testwl::SurfaceId,
+        width: u16,
+        height: u16,
+    ) {
         let data = self.testwl.get_surface_data(surface).unwrap();
         assert!(
             matches!(data.role, Some(testwl::SurfaceRole::Toplevel(_))),
@@ -240,8 +251,8 @@ impl Fixture {
 
         assert_eq!(geometry.x(), 0);
         assert_eq!(geometry.y(), 0);
-        assert_eq!(geometry.width(), 100);
-        assert_eq!(geometry.height(), 100);
+        assert_eq!(geometry.width(), width);
+        assert_eq!(geometry.height(), height);
     }
 
     #[track_caller]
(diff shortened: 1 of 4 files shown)


tests/integration.rs
En-En-Code, line 245: The (100, 100) here are `width` and `height` parameters, so I would expect you need to update them too. But then to get the following asserts to work you need to also pass a `scale` to this function. Ugly.

Untested thoughts: maybe most of this logic could be moved to a `Fixture::move_surface_to_output` that takes `impl Deref` types for `WlSurface` and `WlOutput` and stores the size/scale in those. Or maybe the scale is tracked in Connection since it is global to all outputs (in the current design).

En-En-Code (commented):
What compositor do you use? I could not replicate this issue. No complains about setting the initial `SurfaceScaleFactor` to a more sane value since we easily have it.

5andr0:
> What compositor do you use? I could not replicate this issue. No complains about setting the initial `SurfaceScaleFactor` to a more sane value since we easily have it.

https://github.com/Supreeeme/xwayland-satellite/pull/408
This is all fixed in my giga patch (no merge conflicts with all the bug fixes). I appended a dockerfile where you can easily reproduce this bug. I can reproduce this on latest ubuntu / Mutter compositor.

tests/integration.rs
awsms: I'm on Hyprland, which compositor do you use? This bug makes Ghidra unusable (all the plugins window gets larger & larger, just like in my video repro with ares-emu), the patch works fine!

En-En-Code: Mostly Niri, although recently I have been in Sway. I have reasonable configs to drive Hyprland and i3 (reference X window manager). Helps a lot since compositor-specific bugs can be our fault still.

My blocker was not the fix; it was a concern that the changes to the integration test make it assert the wrong thing. You need not change any code outside of the tests at this point.

En-En-Code: Okay, I finally got back to this. The stuff I suggested doesn't work for various reasons, so a `scale` parameter did need to be added. Fortunately, the `width` and `height` parameters can be fetched from X and so are redundant. The changes the `forced_1x_scale` an additional regression test. Those changes are in a PR to your branch.

Supreeeme:
Sorry for like, ignoring this. This looks like something I've intended to fix for a while. Looks good to me. If it looks good to @En-En-Code I'll merge.

Issue #284, opened 2025-11-22
Scaling breaks on large geometry changes


njdom24:
### Operating System

NixOS Unstable

### Wayland Compositor

Sway

### Describe the issue.

**Please attach a log if applicable.**

[satellte_debug.txt](https://github.com/user-attachments/files/23691298/satellte_debug.txt)

Apps tend to scale too small on geometry changes.

### Steps to reproduce

Happens **100% reliably** when opening Steam and floating it.
Even when unfloating, the window's scaling remains too small, and the app only sees cursor events roughly equal to half the width and height of the actual window. The only way to get it out of this state is to:

- Float the window again
- Enable fullscreen
- Disable floating (no visual impact yet, but I found it's required)
- Disable fullscreen

This also happens often (not 100%) when opening a terminal and tiling it alongside Steam, then killing the terminal, making Steam suddenly resize.

I have 2 displays from left to right:

- **DP-1:** 3840x2160 scale 1.5
- **DP-2:** 1920x1080 scale 1

### Please confirm you have reproduced this on the latest commit.

- [x] I have reproduced this on the latest commit

njdom24:
Still seeing the same behavior with both monitors set to a 1.5 scale. I'm seeing a lot of `wl_output@12[0] scale: 2` and `Using new scale 2` in the output, when I have nothing with a scale of 2 at all.

**Edit:** Also seeing this issue on the `unscaled-dpi` branch

Supreeeme:
What exactly does "scaling breaks" mean? Is it just the cursor behavior or something visual as well?

njdom24:
Apologies, I should have recorded a demo. I'm away for the holidays for the next ~week, but the symptoms are:

- Cursor inputs are only received in the upper-left quadrant of the window (but are not offset from the content)
- The window contents appear to be scaled as it should be if my monitor has 2x scaling set. Since I run 1.5x scaling, the window contents are shrunk down too small.

I'll try to get a remote desktop going to get a visual.

Supreeeme:
Hm, and the scaling only breaks when you float the window? It works fine otherwise?

njdom24:
> Hm, and the scaling only breaks when you float the window? It works fine otherwise?

[...]

Issue #272, opened 2025-11-07
Support resizing from xwl-s CSD


YaLTeR:
**Is your feature request related to a problem? Please describe.**

xwl-s now provides a CSD bar that can be used to drag and close the window, which is great. But another common thing that CSD provides is resize handles for mouse-only resizing.

**Describe the solution you'd like**

Resize handles should be outside the window so as not to interfere with clicking on anything inside the window (see how GTK 4 does it for example). So, following winit and I guess libdecor, xwl-s could add small transparent subsurfaces along the four edges that catch pointer input, change the pointer shape to resize handles, and do window resizing when dragged. Also they should go away when the window is in a maximized or tiled state I guess.

Issue #233, opened 2025-09-02
Popup menus appear off-screen on multi-monitor setup


yunxi177:
### Operating System

Arch Linux

### Wayland Compositor

Niri

### Describe the issue.

Description:
On a multi-monitor setup, popup menus (e.g., right-click context menus) in Xwayland applications appear at the wrong position. Logs show coordinates like -1920x0 for the popup while the parent window is on the laptop screen at (0,0). This seems caused by xwayland_satellite calculating popup positions without considering the parent window’s screen. As a result, popups are created off-screen and invisible.

Expected behavior:
Popup menus should appear near the parent window on the correct screen, taking into account screen layout and DPI scaling

Example log:
```
configuring wl_surface@44[0] (Window { res_id: 4194457 }): -1920x0, 199x75
Trying to update size hints on wl_surface@44[0], but toplevel role data is missing
```
Main screen: eDP-1 at (0,0)
External display: left of main screen, width 1920

### Please confirm you have reproduced this on the latest commit.

- [ ] I have reproduced this on the latest commit

Supreeeme:
What application are you seeing this in? And can you provide a full log?

yunxi177:
application :aur/wechat-bin
full log:
```
2025-09-02T05:15:52.044Z INFO  xwayland_satellite > Starting xwayland-satellite version v0.7-1-gd759c64-dirty
 2025-09-02T05:15:52.056Z INFO  xwayland_process   > _XSERVTransSocketUNIXCreateListener: ...SocketCreateListener() failed
 2025-09-02T05:15:52.056Z INFO  xwayland_process   > _XSERVTransMakeAllCOTSServerListeners: server already running
 2025-09-02T05:15:52.086Z DEBUG xwayland_satellite::server > Using new scale 1
 2025-09-02T05:15:52.087Z DEBUG xwayland_satellite::server::event > moving wl_output@10[0] to 0x0
 2025-09-02T05:15:52.087Z DEBUG xwayland_satellite::server::event > wl_output@10[0] dimensions: 2240x1400
 2025-09-02T05:15:52.087Z DEBUG xwayland_satellite::server::event > wl_output@10[0] scale: 2
 2025-09-02T05:15:52.087Z DEBUG xwayland_satellite::server::event > moving wl_output@10[0] to 0x0
 2025-09-02T05:15:52.087Z DEBUG xwayland_satellite::server::event > moving wl_output@12[0] to -1920x0
[...]

Issue #373, opened 2026-02-09
NOTICE: Rewrite In Progress


Supreeeme:
If you are a regular peruser of satellite issues, you may or may not have noticed that I have been not handling issues as quickly as before. This is because I am currently in the middle of a rewrite of a chunk of satellite's codebase to use Rust's async support.

## Why?
Satellite's codebase has grown quite a bit and as a result has become a bit spaghetti like in places. A rewrite will allow me to organize it a bit better, and I think using async will allow me to keep it organized.

## What does this mean for me?
As a user, nothing, except that I may not address issues as quickly.
As a contributor, also nothing for now. I'm still accepting and will merge PRs, and I will make sure your changes end up in the rewrite.

## When will this be done?
Eventually! 😀
I'm not sure, but hopefully not too long.

CCing frequent contributors: @GoranKovac @En-En-Code

En-En-Code:
Probably would not be a bad idea to cut a new release soon then, so that we have as much time as we need to fix bugs in the rewrite when it lands.

If anything, I am interested to see what you perceive as spaghetti (I can think of a couple places off the top of my head) and how async will "keep it organized." The `O_NONBLOCK` `WritePipe` hack in selection handling would go away, so there's that I guess.

If you want to know what I currently doing, I am redefining the global output offset so that it aligns with how `xrandr` places  outputs, namely that the top-most monitor's top edge is on the X-axis and the left-most monitor's left edge is on the Y-axis. (for some reason tablet tool behavior in KDE apps relies on this assumption). Will be a bit since a bunch of tests need changing and some new one need writing. One already both discovered a problem I had not considered and failed even after I fixed it because I wrote the test wrong (I suck at TDD I guess lol).

Supreeeme:
> Probably would not be a bad idea to cut a new release soon then

True, I will do that soon.

>  I am interested to see what you perceive as spaghetti

- I don't like how everything is smashed into event.rs and dispatch.rs. This could be fixed without async though.
- I don't like how creation of Wayland objects is completely separated from their event handling. This comes down to the design of wayland-client/server, and I like the current approach I have in the async code
- I don't really like the use of the ECS and pulling things out of thin air, though I think it's better than what was there before, which was just a worse ECS basically
- The way updating surface scale works gives me heartburn. I'm not sure if async will fix this though.

I mostly like how composable async is.

En-En-Code:
[...]

PR #408, opened 2026-03-30
Window Scaling,Dragging,Spawning,Snapping,Freezing,Min-Maximizing,Resizing bug fix giga patch

5andr0:
I decided to merge all of my bug fixes into one branch, because they all had merge conflicts. This way, having them sequential, we avoid merge conflicts and there's no need to resolve any conflicts for you.

I will do another donation if you implement this as a thank you for taking the time to read all this.

#  Giga Patch

## Short Bug List

- Startup position bug: some X11 windows were exported to Wayland too early at `0,0`, which could place them on the wrong monitor before their real startup geometry arrived. See https://github.com/Supreeeme/xwayland-satellite/issues/233
- Cross-monitor drift bug: explicitly positioned windows could be shifted again on later `wl_surface.enter` events, causing whole-monitor jumps, drag instability, and focus/output churn.
- Faulty scaling bug: xwayland-satellite was trying to rescale X11 windows itself during output changes, causing shrink/grow loops and broken interaction on mixed-DPI layouts. See https://github.com/Supreeeme/xwayland-satellite/issues/284
- Frameless drag bug: custom-decorated X11 apps that drag by sending mapped `ConfigureRequest` position updates were not translated into a compositor-side interactive move.
- Snap cutoff bug: snapped X11 windows on scaled outputs could negotiate stale or oversized constraints, so the first snap could be too wide and get cut off.
- Minimize/maximize support: add missing X11 minimize/maximize handling, plus capability-aware action exposure.
- In environments without Wayland compositor SSD support (GNOME mutter), xwayland-satellite provides window decorations. However, server-side decorated X11 windows were not given interactive resize regions, making edge resizing unavailable. See https://github.com/Supreeeme/xwayland-satellite/issues/272
- Cursor theme bug: the X11 default cursor theme ignored `XCURSOR_THEME`, causing the root and fallback resize cursors to use the wrong theme. See https://github.com/Supreeeme/xwayland-satellite/issues/421
- Fallback decorations lacked GTK3-style light/dark appearance, color-scheme updates, fades and a proper window shadow.

## Reproduction

```sh
git clone https://github.com/5andr0/xwayland-satellite.git
cd xwayland-satellite
git checkout docker_test
cd docker
# Configure your .env for the correct wayland user, display and runtime paths
./configure.sh

# Builds all branch binaries once, no rebuild needed. Only switch your env vars

# TEST_SSD=1 -> request x11 server side decorations (titlebar, resize drag area). 
# This test is meant for wl compositors without SSD support (like GNOME mutter) 
# when xwayland satellite is in charge of the decorations

# Beyond broken on multi monitor with different scales, dragging across monitors
# TEST_SSD=1 to have working window dragging
MODE=upstream TEST_SSD=1 docker compose up

# Same with a non qt app
MODE=upstream TEST_SSD=1 QT=0 docker compose up

# Dragging doesn't work with custom decorated qt windows
# Snapping to right doesn't work on different scale monitor, especially in portrait mode
# It cuts off the right part of the window, also with titlebar
MODE=upstream TEST_SSD=0 docker compose up

# No window size changes with host xwayland socket:
# Host already properly takes care of independent pixel scaling with different monitor scales
MODE=host_xwl TEST_SSD=0 docker compose up
MODE=host_xwl TEST_SSD=0 QT=0 docker compose up

# Works properly with this PR:
MODE=main TEST_SSD=1 docker compose up
MODE=main TEST_SSD=1 QT=0 docker compose up
```

Freezes/Unresponsiveness can only be seen without the 6032cc01003ab4d4ef1af463faf52ceb03a34990 commit on mixed scale multi monitor, but I don't have a test build for that.

### Problem

Some X11 clients first map at a provisional `0,0` position and only send their real startup geometry afterward. Exporting those windows to Wayland too early lets the compositor classify them on the wrong output.

### Change

Initial toplevel creation is deferred only for the narrow startup case: origin-mapped, non-popup, non-explicitly server-decorated windows. The window is released once post-map geometry arrives, or by the existing timeout fallback, while preserving pre-role surface and buffer state.

### Problem

After the initial placement fix, later `wl_surface.enter` events could still apply another output-origin offset to already-correct X11 coordinates. That caused monitor-sized jumps, drift, drag instability, and focus/output churn.

### Change

The server now tracks whether the output offset has already been initialized. Later enter events update output bookkeeping without rewriting mapped non-popup toplevel coordinates, and the focused-window path no longer resets RandR primary output on every border crossing.

### Problem

The previous scaling path picked the smaller scale on mixed-DPI setups, which made X11 window content blurry on scaled monitors. It also reacted to output changes by mutating window scale and geometry, so monitor-crossing drags could race compositor configures and create shrink/grow loops.

### Change

The dynamic output-enter rescale path is removed. xwayland-satellite now uses one stable X11 scale based on the highest monitor scale, so windows stay sharp on scaled outputs and can cross monitor borders without being resized on every enter event.

Because many X11 GTK apps only support integer scaling, and xwayland-satellite cannot reliably know whether a client supports fractional scaling through `Xft.dpi`, the selected scale is rounded up to the next integer. That integer scale is published through XSettings, and Xwayland output geometry is republished before queued configure requests.

### Problem

Frameless or custom-decorated X11 apps often drag by sending mapped toplevel `ConfigureRequest` X/Y updates. Those were previously forwarded as raw geometry changes instead of starting a compositor-side move, which made custom-decorated X11 windows fail to drag correctly.

### Change

Click-backed, mapped, non-popup, position-only `ConfigureRequest` traffic is translated into `xdg_toplevel.move`. Repeated requests for the same click serial are deduplicated, and the request mask is respected so resize traffic stays on the resize path.

### Problem

On mixed-DPI layouts, snapped windows could negotiate stale or oversized constraints. The first snap could be too wide for the target tile, especially on scaled or portrait outputs.

### Change

Logical output size is tracked separately from raw mode size. Snapped, tiled, maximized, and fullscreen constraints are clamped against logical output bounds, with size limits derived from the active output's effective constraint scale and committed early enough for the first snap.

### Problem

X11 minimize and maximize requests were not fully bridged to Wayland, and reflected X11 state did not consistently match compositor state. The fallback decorations also lacked complete minimize/maximize actions.

### Change

The bridge now handles `WM_CHANGE_STATE` minimize requests and `_NET_WM_STATE_MAXIMIZED_{VERT,HORZ}` maximize/restore requests. Minimized and maximized state is reflected back to X11, and the fallback decoration controls use the same bridge actions.

### Problem

X11 clients and fallback decorations did not know which window actions the Wayland compositor actually supports. That meant actions like minimize, maximize, or fullscreen could be advertised or shown even when the compositor did not expose them for the toplevel.

### Change

`xdg_toplevel::WmCapabilities` is now parsed and reflected into `_NET_WM_ALLOWED_ACTIONS`. Move, resize, and close remain available by default, while minimize, maximize, and fullscreen are only advertised when the compositor reports support for them.

The fallback decoration controls use the same capability state, so unsupported minimize/maximize buttons are not shown.

### Problem

Server-side decorated X11 windows could express resize intent through mapped `ConfigureRequest` geometry changes, but xwayland-satellite treated that as plain reconfiguration instead of compositor-side interactive resize.

### Change

Resize-style configure traffic is classified as move, resize, or locked resize. The bridge reconstructs the requested rectangle, infers the edge or corner, starts `xdg_toplevel.resize` with the stored click serial, and deduplicates repeated requests. SSD frames also get an invisible resize margin, disabled for maximized, fullscreen, fixed-size, or non-resizable windows.

### Problem

xwayland-satellite loads the root cursor through `xcb-util-cursor`, which does not use `XCURSOR_THEME` directly. It looks for `Xcursor.theme` in the X11 root `RESOURCE_MANAGER` instead, so setting `XCURSOR_THEME` alone did not change the loaded cursor theme. `XCURSOR_SIZE` was already respected by the cursor loader; the missing piece was the theme resource.

### Change

Before creating the XCB cursor context, xwayland-satellite now mirrors `XCURSOR_THEME` into `RESOURCE_MANAGER` as `Xcursor.theme`. It also mirrors `XCURSOR_SIZE` as `Xcursor.size` for consistency with the same X resource path.

### Problem

Only in GNOME environments, compositor-provided server-side decorations are not available, so xwayland-satellite has to draw the decorations itself. So why not match the decoration style to a GNOME look like GTK3?

### Change

The fallback titlebar is replaced with GTK3-style decorations: title text (bold), window controls, focus and hover states, resize hit regions, and a softer shadow. Decorations use light or dark palettes from the XDG desktop portal color-scheme setting, fall back to `COLOR_SCHEME` when D-Bus is unavailable, and redraw when the portal reports changes. No preference/default color-scheme values are treated as light.

Rounded corners and borders are adjusted for maximized or tiled windows so snapped windows stretch cleanly instead of keeping floating-window corner treatment.


Supreeeme:
Sorry for the delay in responding to this. I'm going to avoid reviewing this until after I finish the rewrite (#373), just because it's such a large PR. Also, for future reference, while I appreciate the PR, I would appreciate if you did not drop the overly long LLM explanation into the PR description, because it makes me really not to read it. Smaller and more focused PRs are also easier for me to review.

5andr0:
> Sorry for the delay in responding to this. I'm going to avoid reviewing this until after I finish the rewrite (#373), just because it's such a large PR. Also, for future reference, while I appreciate the PR, I would appreciate if you did not drop the overly long LLM explanation into the PR description, because it makes me really not to read it. Smaller and more focused PRs are also easier for me to review.

Thanks for the reply. I don't mind the delay, now I had time to thoroughly test this. There's an overhaul coming this weekend, mainly scaling related.
Sorry about not doing smaller PRs, but I needed all the patches for myself and I couldn't separate them, because of a billion merge conflicts, which weren't easy to solve.
Commit summary sections reduced to bare minimum info. Feel free to ask for more details.

I refactored everything and extensively tested my dockerized apps, which now behave exactly like running natively.

Added changes:

**Scaling behavior on mixed scale monitors follows gnome:**
- If fractional scaling is disabled, the apps change scaling raster and physical window size triggered by compositor when the window is (mostly) fully moved to another monitor (not on each monitor border crossing, which broke the origin implementation completely)
- If fractional scaling is enabled, the compositor takes care of translating logical to physical pixels without ever triggering scaling changes on monitor crosses, but we have to render with the highest scale to achieve sharp rendering on higher scaled monitors. Not like the current implementation of rendering with the lowest scale
- Note: Gnome 50 even removed the option for disabling fractional scaling

**Server side decorations:**
- Gnome is the only environment without server side decorations, so we might as well draw the decorations in gnome gtk3 style (no dependencies added). I also added an env option for DARK/LIGHT mode

**XCURSOR_THEME is now honored correctly**

I really would love to see those fixes integrated, so I don't have to maintain my own fork.  
After your rewrite, I could post single PRs, but we should do this coordinated. Just tell me how I can be of help. Maybe get into contact somehow

awsms:
I don't think it's realistic to expect the maintainers to review +9000 LOC changes entirely written by an LLM, when the project itself is ~16kLOC and being rewritten.....

5andr0:
> I don't think it's realistic to expect the maintainers to review +9000 LOC changes entirely written by an LLM, when the project itself is ~16kLOC and being rewritten.....

After the rewrite I'll try to apply these patches to the new version, when still needed. In the meantime, everyone who encounters these bugs, can use my forks main branch.

Regarding LLMs, this is where things are heading 🤷🏻‍♂️
Without it, noone would waste the time to fix all that. 
And with it, we even have ai code/PR review for maintainers, which also checks for bugs. 
I'd rather embrace it - why would you waste time not writing code with LLM, except for critical stuff?

Most important part here is to check if these changes work or break anything. I supervised the LLM writing these patches over the course of 1 month and poured a lot of time into it, to test every use case, pick the best solution and steer it in the right direction. I even tested on different distros with KDE, latest gnome 50 and so on.
But I needed it for myself anyway, so I've been using it heavily with desktop apps of all sorts and so far everything works like a charm with no difference to native xwayland running apps

awsms:
that's not what I meant, would have been better if you had your own "fork master" branch that copies upstream master + merged your patches in it, and separate branches & PR for each bug fix; rather than "Window Scaling,Dragging,Spawning,Snapping,Freezing,Min-Maximizing,Resizing bug fix giga patch that's 9k changes" lol

5andr0:
> that's not what I meant, would have been better if you had your own "fork master" branch that copies upstream master + merged your patches in it, and separate branches & PR for each bug fix; rather than "Window Scaling,Dragging,Spawning,Snapping,Freezing,Min-Maximizing,Resizing bug fix giga patch that's 9k changes" lol

It's not that easy, each bug fix depends on the previous bug fix changes. With separate PRs it would've been an absolute merge conflict nightmare with thousands of lines (not even exaggerating) that needed manual conflict resolving or simply not even possible without the previous implementation. No maintainer wants to do conflict resolving! 
What I can do after the rewrite is trickle in PR by PR, wait until the first one is merged, then work on the next. 
It'll take ages and back then I needed a fully working fork with everything in it.

This PR is not meant as a "merge please", I just wanted to provide my solution to all the bugs and a working fork for anyone who's struggeling with these bugs like I did. 
I offered the maintainer to get in touch to find a way how to best work together and implement these patches. Let's see how much there is left to fix after the rewrite.
Just wanting to help here

shibahaczix:
We should ban AI pull requests.

PR #410, opened 2026-04-01
Translate X11 urgency hints to Wayland

taladar:
Some applications set the X11 urgency hint to request attention. Wayland has a similar mechanism but XWayland did not translate one into the other. This PR adds that feature.

I have tested this with Pidgin 2.x with the Message Notification plugin with the 'Set window manager "URGENT" hint' option on the niri compositor and it seems to work as intended.

### DIFF
--- a/src/server/mod.rs
+++ b/src/server/mod.rs
@@ -951,6 +951,36 @@ impl<S: X11Selection + 'static> InnerServerState<S> {
         };
 
         self.world.get::<&mut WindowData>(id).unwrap().attrs.group = hints.window_group;
+
+        if hints.urgent {
+            self.request_attention(window);
+        }
+    }
+
+    pub fn request_attention(&mut self, window: x::Window) {
+        let Some(activation_state) = self.activation_state.as_ref() else {
+            return;
+        };
+
+        let app_id = self
+            .windows
+            .get(&window)
+            .copied()
+            .and_then(|id| self.world.entity(id).ok())
+            .and_then(|e| e.get::<&WindowData>())
+            .and_then(|d| d.attrs.class.clone());
+
+        activation_state.request_token_with_data(
+            &self.qh,
+            clientside::ActivationData::new(
+                window,
+                smithay_client_toolkit::activation::RequestData {
+                    app_id,
+                    seat_and_serial: None,
+                    surface: None,
+                },
+            ),
+        );
     }
 
     pub fn set_size_hints(&mut self, window: x::Window, hints: WmNormalHints) {
--- a/src/server/selection.rs
+++ b/src/server/selection.rs
@@ -40,12 +40,17 @@ impl<S: X11Selection> SelectionStates<S> {
     }
 
     pub fn seat_created(&mut self, qh: &QueueHandle<MyWorld>, seat: &WlSeat) {
+        // Track a data device per seat rather than a single one: a compositor may expose more than
+        // one seat (e.g. niri's keyboard-only "injector" seat), and a later seat must not clobber
+        // the device for the seat the user actually copies/pastes on.
         if let Some(c) = &mut self.clipboard {
-            c.device = Some(c.manager.get_data_device(qh, seat));
+            c.devices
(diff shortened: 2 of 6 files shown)


src/server/mod.rs
```diff
@@ -951,6 +951,36 @@ impl<S: X11Selection + 'static> InnerServerState<S> {
...
+
+        activation_state.request_token_with_data(
+            &self.qh,
+            clientside::ActivationData::new(
+                window,
+                smithay_client_toolkit::activation::RequestData {
+                    app_id,
+                    seat_and_serial: None,
+                    surface: None,
+                },
```
Supreeeme, line 981: Do you know if these params are needed on other compositors? What compositor have you tested this on? It would be simple enough to grab these if necessary in any case.

taladar: As far as I know seat_and_serial None and surface None is what distinguishes an urgency hint from a request to be raised and get focus but I am relatively new to Wayland myself.

En-En-Code: My thoughts after reading up and testing stuff (sorry it's a lot):

Making an [`xdg_activation_v1::activate`](https://wayland.app/protocols/xdg-activation-v1#xdg_activation_v1:request:activate) request with an invalid token is allowed, though "compositors *can* ignore unknown activation tokens when an invalid token is passed" (emphasis mine). This openness in the protocol allows compositors to treat an invalid token as a "weak" focus request, usually displayed as a hint. What compositors do with a valid token is also open-ended (e.g. Niri won't focus a window scrolled offscreen, Sway has `focus_on_window_activation` to make it user-configurable, etc.). There's a [great discussion in `wayland-protocols`](https://gitlab.freedesktop.org/wayland/wayland-protocols/-/work_items/150) about this second, undocumented use of activation and how Gtk, Qt, etc. map focus and focus hinting from X to Wayland.

Thus, I think the true question being asked here is not "Does passing an invalid token work on various compositors?"—it's "Should UrgencyHint being set map to an `xdg_activation::activate` intentionally passing an invalid token?" Tokens with invalid serials is how [Niri does urgency hinting](https://github.com/niri-wm/niri/pull/1319/changes#diff-05613bb57bfb64bc5526311e4c7a2179b5fd6c1290e980949d0197b1a783bee2). Sway also allows invalid tokens to hint it, although `focus_on_window_activation` can treat valid activation token as hints or be disabled. After testing these two compositors with both implementations using valid and invalid tokens, I prefer the behavior of invalid tokens because
1. Niri does not fallback to a hint if the window could not be focused (e.g. on a different workspace, scrolled off-screen, etc.)—it does nothing (though arguably this is a Niri problem).
2. I personally hate my focus being moved by application request. I think it is completely within the rights of UrgencyHint to do so. [Its section in the ICCCM](https://tronche.com/gui/x/icccm/sec-4.html#s-4.1.2.4) allows it to raise an urgent window to the top of the stack in a stacking WM, which is similarly intrusive (This may not be the case for [_NET_WM_STATE_DEMANDS_ATTENTION](https://specifications.freedesktop.org/wm/latest/ar01s05.html#id-1.6.8)).

Now convinced invalid tokens are a (though not necessarily the only) reasonable design decision, I'll comment on the implementation. `request_attention` is mostly a copy of `activate_window`. Initially I thought this should be folded into `activate_window`, changing it slightly to pass an invalid token if explicitly requested or if the valid one could not be obtained. `activate_window`, however, has a bug I just now understand. When opening an X application with no others open, `satellite` gives the following warning message:
```
WARN  xwayland_satellite::server > No last focused toplevel, cannot focus window Window { res_id: Xid }
```
It focuses because of a compositor policy that new windows are always focused rather than correct use of `xdg_activation`. This can be seen in Niri by setting the [`strict-new-window-focus-policy` debug option](https://niri-wm.github.io/niri/Configuration%3A-Debug-Options.html#strict-new-window-focus-policy). With this set, launching the first X application will not focus it. Due to this, I would not want a fallback which only hints (and fixing `activate_window` is outside the scope of this PR), so the current duplication is fine.

Issue #413, opened 2026-04-05
xwayland-satellite panics (exit 101) when all outputs disconnect via KVM switch


ymolists:
### Operating System

NixOS

### Wayland Compositor

niri 25.11

### Describe the issue.

When using a hardware KVM switch, xwayland-satellite panics with exit status 101 the moment both monitors are disconnected. This kills all running X11 clients (e.g. `xeyes`). Niri restarts xwayland-satellite when the monitors reconnect, but by then all X11 apps are dead with:

```
X connection to :0 broken (explicit kill or server shutdown).
```

The crash is 100% reproducible — it happens every time I switch away on the KVM.

## Expected behavior

xwayland-satellite should handle the zero-output state gracefully and keep X11 clients alive until outputs return.

## Process tree showing xwayland-satellite i just build from `sha a879e5e` under niri:

```
procs --tree xwayland 
           PID      User │ TTY   CPU MEM CPU Time │ Command
                         │       [%] [%]          │
 └┬─────── 1        root │       0.0 0.0 00:00:05 │ /run/current-system/systemd/lib/systemd/systemd
  └┬────── 320773   ymo  │       0.0 0.0 00:00:00 │ /nix/store/bigkpra9jw48fip69q4wndsf4kb3d2w9-systemd-258.3/lib/systemd/systemd --user
   ├┬───── 347130   ymo  │       5.4 0.2 00:00:08 │ /nix/store/d857pxvxq9r5lv8xv5w5j29rp434irdb-niri-25.11/bin/niri --session
   │└┬──── 350981   ymo  │       0.0 0.0 00:00:00 │ xwayland-satellite :0 -listenfd 156 -listenfd 113
   │ ├──── 350983   ymo  │       0.0 0.1 00:00:00 │ Xwayland :0 -listenfd 156 -listenfd 113 -rootless -force-xrandr-emulation -wm 7 -displayfd 8 -audit 1 -ve
   │ └──── [350988] ymo  │       0.0 0.0 00:00:00 │ .xwayland-satel
   ├┬───── 347174   ymo  │       0.0 0.0 00:00:00 │ /home/ymo/.local/xwayland-satellite.a879e5e
   │├───── 347177   ymo  │       0.0 0.1 00:00:00 │ Xwayland -rootless -force-xrandr-emulation -wm 7 -displayfd 8
   │└───── [347178] ymo  │       0.0 0.0 00:00:00 │ .xwayland-satel
[...]

PR #406, opened 2026-03-26
fix: crash when surface destroyed during output removal

abcfy2:
When a Wayland output is removed (e.g., screen turns off), surfaces on that output may be destroyed simultaneously. The code previously used .unwrap() on query results that could fail when:
1. The surface entity was despawned (query_one returns Err)
2. The surface's components were removed (get() returns None)

This commit handles both cases gracefully by using if let patterns instead of unwrap(), preventing panics during output removal.

Fixes #405

PR #414, opened 2026-04-07
Log to file when integrated, document running latest build with integration

En-En-Code:
After multiple instances of an integration with Niri relaunching a old version (as opposed to the latest commit/patch build) on their behalf (see comment on #340, #406, #413), I have done two things to reduce user/dev confusion. 
* I added documentation on configuring a `satelllite` integration to run their build
* When integrated with the `-listenfd` flag, it now logs to a file. This log file contains the version of `satellite` as the first line of the log, confirming whether the latest build is being run, and makes reporting `satellite` bugs easier for people who use it integrated with their compositor (for this motivation, logging to a file also ups the log level to Debug).

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,10 +26,14 @@ Some (most?) Java applications may present themselves as a blank screen by defau
 cargo build
 # release build
 cargo build --release
+```
 
-# run - will also build if not already built
+## Running
+```
+# will also build if not already built
 cargo run # --release
 ```
+If `satellite` is integrated into your compositor, the compositor's integration will take precedence over other running instances, even if its `satellite` instance is not running. To test latest commit and/or changes, either disable the integration, configure it to launch your built executable, or install `satellite` to the currently-configured location (likely `/usr/bin/xwayland-satellite`) and kill all running `satellite`. In all but the last case, you will need to restart the compositor for the changes to take effect.
 
 ## Systemd support
 xwayland-satellite can be built with systemd support - simply add `-F systemd` to your build command - i.e. `cargo build --release -F systemd`.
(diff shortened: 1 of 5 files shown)


En-En-Code:
Okay, some big changes, but I think I finally have this where I want it. When running as an integration, it logs `stderr` (including any panic messages) to a timestamped file, and set logging to be more verbose when built in debug mode. Also restructured the README so related things are next to each other since I promised YaLTeR that.

README.md
```diff
@@ -4,38 +4,57 @@ This is particularly useful for compositors that (understandably) do not want to
 
 Found a bug? [Open a bug report.](https://github.com/Supreeeme/xwayland-satellite/issues/new?template=bug_report.yaml)
 
-Need help troubleshooting, or have some other general question? [Ask on GitHub Discussions.](https://github.com/Supreeeme/xwayland-satellite/discussions)
+## Wayland protocols used
```
Supreeeme, line 7: I feel like this section should go near the `Compositor integration` section.

src/main.rs
```diff
@@ -246,3 +244,38 @@ fn parse_args() -> RealData {
 
     data
 }
+
+fn init_logger(integrated: bool) {
+    || -> Option<()> {
+        if !integrated {
+            return None;
+        }
```
Supreeeme, line 252: This construct is kinda weird. You could just create a lambda with everything after the `integrated` check and only run it if `integrated` is true.

Supreeeme (commented):
One of the alternative suggestions was to log to journalctl. Any thoughts on that approach instead?

En-En-Code:
> One of the alternative suggestions was to log to journalctl. Any thoughts on that approach instead?

`journald` aggregates messages from `syslog` calls, so that would be the tool I would reach for first, benefit being not depending on anything specific to `systemd`.

PR #395, opened 2026-03-05
Check _WINE_HWND_STYLE flags for POPUP

GoranKovac:
Steam games provide this Atom. Since `_MOTIF_WM_HINTS` can fail on them we check _WINE_HWND_STYLE` flags: https://learn.microsoft.com/en-us/windows/win32/winmsg/window-styles
https://learn.microsoft.com/en-us/windows/win32/winmsg/extended-window-styles

We determine by `WS_POPUP, WS_CAPTION, WS_THICKFRAME` + `WS_EX_APPWINDOW, WS_EX_TOOLWINDOW` is the wine window a popup.

Hopefully fixes https://github.com/Supreeeme/xwayland-satellite/issues/392
Fixes also: https://github.com/Supreeeme/xwayland-satellite/issues/392#issuecomment-4006218161

Setting as draft since it needs testing + more work (breaks other steam apps).

EDIT: Latest commit tested across these games/apps no regressions seen:
image

NOTE: This fix affects only games run thru steam/proton

PR #417, opened 2026-04-15
fix: improve window classification to prevent games being embedded as popups

### LINKED ISSUE #392: Forza Horizon 5 starts "embedded" into steam
### Operating System

Arch Linux

### Wayland Compositor

Niri

### Describe the issue.

Image

### Steps to reproduce

RimuruW:
## Summary

Fix games like Forza Horizon 5 and Cyberpunk 2077 being incorrectly classified as popups and embedded into Steam's window.

I've been experiencing this issue myself while trying to play Cyberpunk 2077 on Steam with Arch Linux + Niri. After waiting over a month without being able to play, I decided to take a shot at fixing it myself.

I referenced the approach in #395, though I ended up taking a slightly different direction — focusing on improving the window classification logic itself rather than adding Wine-specific checks.

## Root Cause

Games often set X11 properties (like `_NET_WM_WINDOW_TYPE_NORMAL`) **after** mapping their windows. The previous heuristics could misclassify windows with incomplete property information as popups, causing them to be embedded into the parent window (Steam).

## Changes

- Replace boolean `is_popup` with typed `WindowKind` enum for clearer semantics
- Introduce `PopupKind` with strong/weak classification to distinguish explicit popup evidence from heuristic guesses
- Only classify as popup when **strong evidence** exists:
  - `override_redirect`
  - Explicit popup window types (menu, tooltip, combo, etc.)
  - Motif no-functions hint
- Default ambiguous windows to toplevel (safer failure mode — a misplaced toplevel is far less disruptive than a game window embedded as popup)
- Re-evaluate window kind when relevant properties change after map
- Weak popup heuristics (e.g. `_NET_WM_STATE_SKIP_TASKBAR`) no longer auto-parent to last focused window

## Test Plan

- [x] All existing tests pass (82 tests)
- [x] `cargo clippy` clean
- [x] Manually verified Cyberpunk 2077 launches correctly as a standalone window on Arch Linux + Niri

---

If there's anything that needs to be changed or if you have concerns about this approach, please let me know — I'm happy to iterate on it.

Fixes #392

### DIFF
--- a/src/server/mod.rs
+++ b/src/server/mod.rs
@@ -7,7 +7,10 @@ pub(crate) mod selection;
 mod tests;
 
 use self::event::*;
-use crate::xstate::{Decorations, MoveResizeDirection, WindowDims, WmHints, WmName, WmNormalHints};
+use crate::xstate::{
+    Decorations, MoveResizeDirection, PopupKind, WindowDims, WindowKind, WmHints, WmName,
+    WmNormalHints,
+};
 use crate::{X11Selection, XConnection, timespec_from_millis};
 use clientside::MyWorld;
 use decoration::{DecorationsData, DecorationsDataSatellite};
@@ -97,9 +100,9 @@ where
     u32::from(wenum).try_into().unwrap()
 }
 
-#[derive(Default, Debug)]
+#[derive(Debug)]
 struct WindowAttributes {
-    is_popup: bool,
+    window_kind: WindowKind,
     dims: WindowDims,
     size_hints: Option<WmNormalHints>,
     title: Option<WmName>,
@@ -107,6 +110,21 @@ struct WindowAttributes {
     group: Option<x::Window>,
     decorations: Option<Decorations>,
     transient_for: Option<x::Window>,
+}
+
+impl Default for WindowAttributes {
+    fn default() -> Self {
+        Self {
+            window_kind: WindowKind::Toplevel(crate::xstate::ToplevelReason::Default),
+            dims: WindowDims::default(),
+            size_hints: None,
+            title: None,
+            class: None,
+            group: None,
+            decorations: None,
+            transient_for: None,
+        }
+    }
 }
 
 #[derive(Debug, Default, PartialEq, Eq, Copy, Clone)]
@@ -128,7 +146,13 @@ impl WindowData {
         Self {
             mapped: false,
             attrs: WindowAttributes {
-                is_popup: override_redirect,
+                window_kind: if override_redirect {
+                    WindowKind::Popup(PopupKind::strong(
+                        crate::xstate::PopupReason::OverrideRedirect,
+                    ))
+                } else {
+                    WindowKind::Toplevel(crate::xstate::ToplevelReason::Default)
+                },
                 dims,
                 ..Default::default()
             },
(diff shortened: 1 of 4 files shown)


GoranKovac:
This unfortunately breaks all yabridge popups, now they spawn as toplevel.
It did not just affected `_NET_WM_WINDOW_TYPE_NORMAL` but also `TYPE_POPUP`.

image

RimuruW:
> This unfortunately breaks all yabridge popups, now they spawn as toplevel. It did not just affected `_NET_WM_WINDOW_TYPE_NORMAL` but also `TYPE_POPUP`.
> 
> image

Sorry, let me look into this and figure out how to fix it.

En-En-Code:
Passing the integration tests by regressing a check with no justification will cause regressions. Considering Yabridge was the original source of the `WM_HINTS` popup heuristic (#328, #355) that has been causing problems, I do not think there's a pretty solution to give both Yabridge and the Steam apps what they want.

Personal opinion: I think we have hit the point where the `WM_HINTS` heuristic has had too large a blast radius and caused more trouble than it is worth. This is messed up in my rewrite [when I mislabeled `PixelComposer` as a poup](https://github.com/Supreeeme/xwayland-satellite/pull/390/changes/83eac5713d40a5099b250403d98089de50f1109d#diff-f0ca6eb7a6e8ff0789737d6a04585ffd307f4f49a8aef1a26a6e26fbc9df5b86R297-R312) (see [comment where `PixelComposer` is a top-level](https://github.com/Supreeeme/xwayland-satellite/pull/328#issuecomment-3754962392)). `PixelComposer` is similar to other Steam apps that I do not see a distinguishing characteristic in. I do not hate removing the `WM_HINTS` check even if it regresses Yabridge.

The problem with both of you two's approaches is that they try to stack rules on rules hoping it all works out. I have concluded Yabridge is an *exception* and should be treated exceptionally (see #315 for prior art). The core pop-up checks are kept simple and edge cases get edge case code. You could accomplish this with almost nothing but code *removal*. +260 lines for this speak volumes about how easy overabstratcting is relative to understanding the actual problem.

GoranKovac:
> Personal opinion: I think we have hit the point where the `WM_HINTS` heuristic has had too large a blast radius and caused more trouble than it is worth. This is messed up in my rewrite [when I mislabeled `PixelComposer` as a poup](https://github.com/Supreeeme/xwayland-satellite/pull/390/changes/83eac5713d40a5099b250403d98089de50f1109d#diff-f0ca6eb7a6e8ff0789737d6a04585ffd307f4f49a8aef1a26a6e26fbc9df5b86R297-R312) (see [comment where `PixelComposer` is a top-level](https://github.com/Supreeeme/xwayland-satellite/pull/328#issuecomment-3754962392)). `PixelComposer` is similar to other Steam apps that I do not see a distinguishing characteristic in. I do not hate removing the `WM_HINTS` check even if it regresses Yabridge.
> 

I agree, it takes away too much from day to day operations. DAWs etc are not something majority of people use every day and should not be treated as priority. So yeah I'm also for removing it.

Will be missed but  for my needs since I'm only using it for DAW (I'm audio engineer) local version is enough for me.

RimuruW:
After looking into this more carefully, I agree that the `WM_HINTS` heuristic is the root of the problem rather than something worth patching around.

My thinking is to remove the broad `WM_HINTS`/MOTIF check from core classification entirely. If yabridge support is worth preserving, it could be replaced with an explicit `WM_CLASS` exact-match for known yabridge process names (`yabridge-host.exe`, `yabridge-bridge.exe`), following the same pattern as #315. That way the core logic stays simple, the blast radius goes to zero, and the yabridge case gets a small allowlist with a comment pointing to upstream. The net diff would be mostly removals.

I'm also fine with just removing the heuristic and accepting that yabridge regresses for now. That's a cleaner result, and it would be more honest about what the X11 properties actually communicate. Yabridge could fix this on its side by setting correct window types, or it could be revisited in a separate issue.

PR #424, opened 2026-04-24
fix: remove WM_HINTS popup heuristic

### LINKED ISSUE #392: Forza Horizon 5 starts "embedded" into steam

RimuruW:
## Summary

Remove the broad `WM_HINTS` popup heuristic for `_NET_WM_WINDOW_TYPE_NORMAL` windows.

Windows with `WM_HINTS` input focus set to false and client-side/no Motif decorations are not reliably popups. Steam/Proton game windows and launchers can expose the same shape, which causes them to be embedded into the last focused Xwayland window.

This keeps the simpler/stronger popup signals intact:
- `override_redirect`
- explicit popup window types
- Motif no-functions hint
- existing utility/no-decoration behavior

## Notes

This intentionally stops treating the `NORMAL + WM_HINTS input=false + no decorations` property shape as a generic popup signal. If a specific application still needs popup behavior for this shape, it should probably be handled as a narrow exception rather than a core heuristic.

Fixes #392
Supersedes #417

## Test plan

- [x] `cargo test popup_heuristics --test integration` fails before the production change
- [x] `cargo test popup_heuristics --test integration`
- [x] `cargo build --all-targets --profile ci --locked`
- [x] `cargo test --profile ci --locked -- --test-threads 1`
- [x] `cargo fmt --check`
- [x] `cargo clippy --all-targets --profile ci --locked --workspace`

Note: a later local rerun of the full integration suite showed two unrelated-looking flaky failures (`client_init_move`, `wayland_then_x11_clipboard_owner`) that both passed when rerun individually.

### DIFF
--- a/src/xstate/mod.rs
+++ b/src/xstate/mod.rs
@@ -635,7 +635,6 @@ impl XState {
         let class = self.get_wm_class(window);
         let size_hints = self.get_wm_size_hints(window);
         let motif_wm_hints = self.get_motif_wm_hints(window);
-        let wm_hints = self.get_wm_hints(window);
         let mut title = name.resolve()?;
         if title.is_none() {
             title = self.get_wm_name(window).resolve()?;
@@ -650,7 +649,6 @@ impl XState {
         if let Some(hints) = size_hints.resolve()? {
             server_state.set_size_hints(window, hints);
         }
-        let wmhints = wm_hints.resolve()?;
         let motif_hints = motif_wm_hints.resolve()?;
         if let Some(decorations) = motif_hints.as_ref().and_then(|m| m.decorations) {
             server_state.set_win_decorations(window, decorations);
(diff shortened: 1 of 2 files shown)


MusicalArtist12:
Looks like this fixes #426

RimuruW:
@MusicalArtist12 Thanks for testing. Interesting that the main UE window now works while the notification toasts break. The two probably expose different property shapes, and the old heuristic was catching both: the toasts (correctly, as it turns out) along with the Steam/Proton false positives this PR is trying to remove.

I'll look into it this week. Rough plan:

1. Reproduce the broken notification locally and run `xprop` on the toast window while it's mapped, so I can see which combination of `WM_HINTS`, `_NET_WM_WINDOW_TYPE`, `_MOTIF_WM_HINTS` and `override_redirect` it actually sets.
2. Find a narrower signal, ideally something the toast has but a fullscreen game doesn't, instead of the broad property shape this PR removes.

If you can share which UE title and what in-game action triggers the notification, that would help a lot. An `xprop` dump of the affected window while it's visible would be even better.

Keeping this in draft until the regression is sorted out.

Issue #189, opened 2025-06-25
Star Citizen Launcher Crash


Fireye04:
### Operating System

Arch Linux

### Wayland Compositor

Niri

### Describe the issue.

A Clean Star Citizen installation results in a partial launcher crash in which the process continues in the background (music playing and everything) but the launcher UI crashes after the login step.

Functions perfectly fine when run on sway, or on niri through xwayland-run or gamescope.

### Steps to reproduce

Follow the quick start guide laid out [here](https://github.com/starcitizen-lug/knowledge-base/wiki/Quick-Start-Guide), launch `./Games/star-citizen/sc-launch.sh`, login, and observe crash. The same installation can be tested in a similar manner on other xwayland solutions and observed to function normally.

Only the launcher installation is required (not the full 100GB game), however as the crash occurs after login, a free RSI account (which can be acquired [here](https://robertsspaceindustries.com/en/enlist), with a [tempmail](https://temp-mail.org/) address or otherwise) is likely required to reproduce it.

### Please confirm you have reproduced this on the latest commit.

- [x] I have reproduced this on the latest commit

Supreeeme:
Can you run satellite with the env vars `RUST_LOG=debug` and `RUST_BACKTRACE=1` set, and then reproduce the issue and post the log?

Fireye04:
Sure, here it is: 

[sat.log](https://github.com/user-attachments/files/20918452/sat.log)

Supreeeme:
Is this the full log? I guess it's just the UI crashing and not satellite itself. Does anything interesting get printed to logs/stdout when it crashes?

Fireye04:
Here's what I ran to get the log: 

`RUST_BACKTRACE=1 RUST_LOG=debug xwayland-satellite  2>&1 | tee sat.log`

This could be getting the wrong thing though. I'll see if I can dig through the launcher & wine logs as well to see if there's anything there

Fireye04:
Looking at the timing on the satellite output, it stops for a second on this:

```
2025-06-29T05:59:49.904Z DEBUG xwayland_satellite::xstate            > ChangeProperty failed (Window { res_id: 0 }: X(Window(ValueError { response_type: 0, error_code: 3, sequence: 267, bad_value: 0, minor_opcode: 0, major_opcode: 18, pad: 1 }), Some("x::ChangeProperty")))
```

then we get this glob directly as the launcher crashes 
```
[...]

Issue #418, opened 2026-04-16
release and version do not match


bomgar:
The commit that bumps the version to 0.8.1 was in March
https://github.com/Supreeeme/xwayland-satellite/commit/309d8e2a29953f7465dc14c939e2afe4682c0aa9

The release is from February. 

I noticed this because I need this bugfix: https://github.com/Supreeeme/xwayland-satellite/pull/406

So it is not part of the 0.8.1 tag/release. But the version bump in Cargo.toml happened later

Supreeeme:
Yeah, as I noted in #384 I didn't want to rewrite history. Someone else asked for this bugfix as well so I'll do a 0.8.2 release.

thbemme:
Came across the same issue: 
```nix
(~) /nix/store/3lpmx0799dnw8qkzy80gyqfbfjgcf51r-xwayland-satellite-0.8.1/bin/xwayland-satellite -version
0.8.0
```
and my non wayland applications keep crashing when the monitor turns off on Niri (NixOS 26.05).

En-En-Code:
This will have to remain as an idiosyncracy of 0.8.1, but 0.8.2 has the correct version number.

Issue #383, opened 2026-02-21
DaVinci Resolve - Transcription window trapped in main window + Erratic movement


ENBYSS:
### Operating System

EndeavourOS

### Wayland Compositor

Niri (git)

#### Update

The "trapping" is intended behaviour since it is treated as a popup and handled accordingly. The main issue is now the erratic movement for the popup.

Here's what the movement looks like on the latest commit - showing extremely amplified movements:
https://github.com/user-attachments/assets/4537ee5c-9598-468b-a813-dec4ad619539

And here's what it looks like on the package - showing highly erratic movement that leads to the server crashing. Warning for photosensitive epilepsy, just incase:
https://github.com/user-attachments/assets/f813b971-bec0-4127-8f1f-4d60ccf9d186

#### Old Description

If I choose a file to try and see its transcription, the window does pop up, but it is trapped within the main window. By resizing, it can appear outside - but the moment you try and drag it, it remains trapped. Also, dragging it makes the window move very weirdly --- amplifying any motions, so dragging it a bit left just moves the window to the left-most position.

I'm using the latest version of Davinci Resolve, 20.3.1.

Here is the `xprop` for the main Resolve window
```
WM_STATE(WM_STATE):
		window state: Normal
		icon window: 0x0
_NET_WM_USER_TIME(CARDINAL) = 2134209
_NET_WM_STATE(ATOM) =
_NET_WM_ICON_NAME(UTF8_STRING) =
_NET_WM_ICON(CARDINAL) = 	Icon (48 x 48):
XdndAware(ATOM) = BITMAP
WM_NAME(STRING) = "DaVinci Resolve Studio - heaven-does-not-respond"
_NET_WM_NAME(UTF8_STRING) = "DaVinci Resolve Studio - heaven-does-not-respond"
_MOTIF_WM_HINTS(_MOTIF_WM_HINTS) = 0x2, 0x1, 0x0, 0x0, 0x0
_NET_WM_WINDOW_TYPE(ATOM) = _KDE_NET_WM_WINDOW_TYPE_OVERRIDE, _NET_WM_WINDOW_TYPE_NORMAL
_XEMBED_INFO(_XEMBED_INFO) = 0x0, 0x1
WM_CLIENT_LEADER(WINDOW): window id # 0x600008
WM_HINTS(WM_HINTS):
		Client accepts input or input focus: True
		Initial state is Normal State.
		window id # of group leader: 0x600008
[...]

PR #430, opened 2026-05-12
Heuristics refactor 2

En-En-Code:
Continues #390.
The history was getting a bit bloated, so I rebased it down. Nothing in here should surprise you. Also incorporates the hard-coded exception to Yabridge you approved of in that discussion.


src/xstate/tests.rs
```diff
@@ -0,0 +1,509 @@
...
+        self
+    }
+    fn win_gravity(mut self, gravity: xcb::x::Gravity) -> Self {
+        self.fields[0] |= WmSizeHintsFlags::Gravity.bits();
+        self.fields[17] = gravity as _;
+        self
+    }
+}
+
+mod wrh {
```
Supreeeme: What is `wrh`?

Supreeeme (commented):
Looks good, latest commit caused some conflicts.

src/xstate/tests.rs
```diff
@@ -0,0 +1,509 @@
+#![allow(clippy::needless_update)]
```
Supreeeme: Why is this allowed?

En-En-Code: So that Clippy does not complain about `..Default::default` being used if every field of `WindowRoleHeuristics` is already initialized. My idea behind this was that I wanted adding new fields to `WindowRoleHeuristics` be as easy as possible, since historically, many first-time contributers modify these heuristics.

I feel it is less reasonable to ask those contributers to dig through a bunch of random old issues when I have all of the relevant `xprop` outputs saved on my computer and can just `grep` them and update the test `WindowRoleHeuristics` myself.

zakissimo:
Confirming this fixes a real case for me: Sierra Chart's floating "Chart Trade" window (Wine).

On the current release that window — `override_redirect` + `_NET_WM_WINDOW_TYPE_DIALOG` + fixed size (min==max 722×424) — gets mapped to an `xdg_popup`, so it can't be moved at all: click-dragging its title bar does nothing, niri can't move a popup, and the popup grab thrashes pointer focus (the pointer rapidly enter/leaves the floating surfaces), so Sierra's move-loop never even starts.

On your branch it comes through as a fixed-size toplevel and drags normally — both Sierra's own title-bar drag and niri Super+drag. Exactly the right direction.

src/xstate/tests.rs
Supreeeme: Hm, I find an unneeded `..Default::default` to be a bit confusing personally. If a new field gets added it should be simple enough to add it later to the spots where it's needed. If someone new adds a new field there's probably no reason for that field to be set in existing tests anyway.

En-En-Code: The reason you add new fields to existing tests is for proper regression checking. If someone adds a new field, every application which has that field still needs to return the same heuristic it did previously, so the value of that field needs to be added to them.
The other consequence of always writing `..Default::default()` is that when a new field is added, there are no compiler errors for not initializing every field of `PopupHeuristics`, but I am not opinionated enough to not remove this.

awsms:
praying for this getting ready & merged anytime soon :pray:

Supreeeme:
Sorry for the delay. Thanks for your patience.

En-En-Code:
Full list of issues this closes (correct me if I'm wrong): #280, #307, #383, #392, #438

Other issues which can be closed: #189 (was incidentally fixed by a earlier comment and not noticed for a while, see comments), #418 (new release correctly versioned and tagged, congrats).

PR #431, opened 2026-05-14
selection: bridge via ext-data-control-v1

DzmingLi:
## Symptom

X11 apps that set the clipboard before any of their X11 windows have received Wayland keyboard focus end up with their data source created on the Wayland side but never registered with the compositor — Wayland clients see an empty clipboard.

The triggering case is Electron run with `--ozone-platform=wayland`. Its windows are Wayland-native (so satellite never sees their keyboard focus) but Chromium still keeps the X11 connection open and routes `clipboard.writeText` through it. `xwayland_satellite::xstate::selection::handle_target_list` fires and calls `server_state.set_selection_source`, but `last_kb_serial` is `None` so the `wl_data_device.set_selection` call is silently skipped.

Reproducer: QQ NT on niri 25.08 + xwayland-satellite 0.8.1. `xclip -selection clipboard -o` (X11) returns the copied content, `wl-paste` (Wayland) returns nothing, X11-only apps like WeChat can paste it fine.

## Why an input serial isn't the right answer

`wl_data_device.set_selection` needs a recent input event serial because the protocol gates "background clients changing the clipboard" behind "user just interacted with you". For a clipboard bridge that's the wrong layer — there is no user interaction to point at, and `ext-data-control-v1` was added precisely to give privileged clients (clipboard managers, X11 bridges, …) a serial-free path. One device handles both regular selection and primary selection, and the `set_selection` / `set_primary_selection` requests take no serial.

## Change

Replace satellite's `wl_data_device` + `zwp_primary_selection_device_v1` use with `ext-data-control-v1`. On compositors that don't expose the protocol (Mutter, Weston) the bridge is disabled with a log message — those compositors all have their own built-in Xwayland integration and xwayland-satellite sees little use there.

Mechanically:

- Drop the smithay-client-toolkit `data_device` / `primary_selection` delegates and hand-write the four `Dispatch` impls (manager, device, source, offer).
- Selection slots no longer carry a per-protocol Source/Offer type. Clipboard vs Primary is picked at the device call site (`set_selection` vs `set_primary_selection`).
- Drop the keyboard-serial path on the X11→Wayland side. `last_kb_serial` stays around because xdg_activation still uses it; it's just no longer load-bearing for selection.
- testwl now serves ext-data-control instead of wl_data_device + zwp_primary_selection. The existing selection tests exercise the new wire path unchanged.

## Test

`copy_from_x11_without_x11_focus` (one variant per selection type, six lines in `tests.rs`) reproduces the bug: it calls `set_selection_source` before `create_toplevel` and then asserts the mimes show up. On main it panics with `No selection set on data device` (and `No selection set on primary device` for the primary variant). On this branch it passes, alongside the 48 pre-existing selection tests.

If you'd like to confirm the bug exists on main, the new test in `src/server/tests.rs` is self-contained — cherry-pick that test function plus the two macro lines onto main and `cargo test --lib copy_from_x11_without_x11_focus` will panic with `No selection set on data device` / `No selection set on primary device`.

## Compositor coverage

`ext-data-control-v1` is supported on the compositors that are satellite's actual audience — niri, COSMIC, recent sway / hyprland, and so on. Older wlroots-based versions that only ship `wlr-data-control-unstable-v1` lose the bridge until they pick up `ext-`; a fallback path is straightforward to add in a follow-up if that matters. KDE Plasma, GNOME Mutter, and Weston also don't expose `ext-data-control-v1`, but they all have built-in Xwayland integration of their own, so xwayland-satellite sees little use there in practice.

### DIFF
--- a/src/server/tests.rs
+++ b/src/server/tests.rs
@@ -1388,6 +1388,11 @@ macro_rules! selection_tests {
             }
 
             #[test]
+            fn copy_from_x11_without_x11_focus() {
+                super::copy_from_x11_without_x11_focus::<$selection_type>();
+            }
+
+            #[test]
             fn copy_from_wayland() {
                 super::copy_from_wayland::<$selection_type>();
             }
@@ -1444,6 +1449,26 @@ fn copy_from_x11<T: SelectionTest>() {
         true
     });
     assert_eq!(*mimes, data);
+}
+
+fn copy_from_x11_without_x11_focus<T: SelectionTest>() {
+    let (mut f, comp) = TestFixture::new_with_compositor();
+
+    let mimes = std::rc::Rc::new(vec![testwl::PasteData {
+        mime_type: "text".to_string(),
+        data: b"abc".to_vec(),
+    }]);
+
+    f.satellite.set_selection_source::<T::SelectionType>(&mimes);
+    f.run();
+
+    let win = Window::new(1);
+    let (_surface, _id) = f.create_toplevel(&comp, win);
+
+    let server_mimes = T::mimes(&mut f.testwl);
+    for mime in mimes.iter() {
+        assert!(server_mimes.contains(&mime.mime_type));
+    }
 }
 
 fn copy_from_wayland<T: SelectionTest>() {
--- a/src/server/tests.rs
+++ b/src/server/tests.rs
@@ -1359,11 +1359,26 @@ trait SelectionTest {
         testwl: &mut testwl::Server,
         send_data: impl SendDataForMimeFn,
     ) -> Vec<testwl::PasteData>;
+    fn control_mimes(testwl: &mut testwl::Server) -> Vec<String>;
+    fn control_paste_data(
+        testwl: &mut testwl::Server,
+        send_data: impl SendDataForMimeFn,
+    ) -> Vec<testwl::PasteData>;
+    fn backend(testwl: &testwl::Server) -> Option<testwl::SelectionBackend>;
     fn create_offer(testwl: &mut testwl::Server, data: Vec<testwl::PasteData>);
 }
 
 macro_rules! selection_tests {
-    ($name:ident, $selection_type:ty, $get_mime_fn:ident, $get_paste_data_fn:ident, $create_offer_fn:ident) => {
+    (
+        $name:ident,
+        $selection_type:ty,
+        $get_mime_fn:ident,
+        $get_paste_data_fn:ident,
+        $get_control_mime_fn:ident,
+        $get_control_paste_data_fn:ident,
+        $get_backend_fn:ident,
+        $create_offer_fn:ident
+    ) => {
         impl SelectionTest for $selection_type {
             type SelectionType = $selection_type;
             fn mimes(testwl: &mut testwl::Server) -> Vec<String> {
@@ -1374,6 +1389,18 @@ macro_rules! selection_tests {
                 send_data: impl SendDataForMimeFn,
             ) -> Vec<testwl::PasteData> {
                 testwl.$get_paste_data_fn(send_data)
+            }
+            fn control_mimes(testwl: &mut testwl::Server) -> Vec<String> {
+                testwl.$get_control_mime_fn()
+            }
+            fn control_paste_data(
+                testwl: &mut testwl::Server,
+                send_data: impl SendDataForMimeFn,
+            ) -> Vec<testwl::PasteData> {
+                testwl.$get_control_paste_data_fn(send_data)
+            }
+            fn backend(testwl: &testwl::Server) -> Option<testwl::SelectionBackend> {
+                testwl.$get_backend_fn()
             }
             fn create_offer(testwl: &mut testwl::Server, data: Vec<testwl::PasteData>) {
                 testwl.$create_offer_fn(data);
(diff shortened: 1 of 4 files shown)


src/server/event.rs
```diff
@@ -894,6 +896,9 @@ impl Event for client::wl_keyboard::Event {
                     serial,
                 ));
                 keyboard.key(serial, time, key, convert_wenum(key_state));
+                drop(keyboard);
+                let _ = data;
+                state.flush_pending_selections();
```
En-En-Code: Deferring the clipboard synchronization until a keyboard serial shows up is not an ideal solution. The underlying problem is a Wayland event serial is needed to sync the Wayland clipboard but the X clipboard updating does not obtain such a serial. I will do some research into serials and see what I can come up with.

Also, why `let _ = data;`?

src/server/tests.rs
```diff
@@ -1446,6 +1458,30 @@ fn copy_from_x11<T: SelectionTest>() {
     assert_eq!(*mimes, data);
 }
 
+// X11 sets the selection before any X11 window has received keyboard focus,
+// so `last_kb_serial` is None and `set_selection` must be deferred until a
+// serial arrives. Regression test for clipboard not reaching Wayland when
+// the X11 source is an Electron app using X11 only for clipboard.
+fn copy_from_x11_without_x11_focus<T: SelectionTest>() {
```
En-En-Code, line 1499: This test passes without pulling down the fix, so it does not prove anything.

DzmingLi:
@En-En-Code thanks for the review on the previous version. You were right that deferring `set_selection` until a `wl_keyboard` serial arrives wasn't the right shape of fix — I've rewritten the PR to use `ext-data-control-v1`.

One small thing to double-check on your end if you have a moment: the regression test `copy_from_x11_without_x11_focus` (in `src/server/tests.rs`) reproduces the original bug on `main` for me — both the clipboard and primary variants panic with `No selection set on data device` / `No selection set on primary device` when the test function and its two macro entries are cherry-picked onto `main` and run with `cargo test --lib copy_from_x11_without_x11_focus`. If you had a chance to try it under a different setup last time and saw it pass, I'd love to know what configuration that was — there might be a state ordering I'm missing.

src/server/tests.rs
En-En-Code: My bad, thought I had cleaned the worktree but didn't.

En-En-Code:
Maintain history. Force-pushing should only be necessary if what you pushed was a mistake. I would rather see design history in the back-and-forth phase and squash that down at the end. This goes doubly for your PR comments. Make a new comment instead of expecting me to reread a wall of text for changes.

Though a simple solution, it is not the best. Using the core Wayland clipboard API and the [widely-implemented](https://wayland.app/protocols/primary-selection-unstable-v1#compositor-support) primary selection API give satellite maximum compatability, which this loses. Despite your claims, people DO use this project with both Weston and GNOME Mutter, so completely removing clipboard sync for this small group is not acceptable.

Supreeeme:
Blowing up the existing clipboard support is a bad idea, but supplementing it with `ext-data-control` would be fine in my opinion. It would only have to be used if a client is trying to set the keyboard before it gets focus. I don't know if there's another way to accomplish that without `ext-data-control`.

DzmingLi:
Sorry, I’ve been a bit busy lately and this PR slipped my mind. I reworked this in c512773 so ext-data-control supplements the existing selection support instead of replacing it.

En-En-Code:
> I don't know if there's another way to accomplish that without ext-data-control.

There technically is a clipboard portal in `xdg-desktop-portal-{gnome,kde}`, but I do not think it is worth the code complexity. I honestly do not know why these portals exist instead of using the actually interoperable Wayland protocol. All they do is make me better understand the "desktop portals suck" and "Wayland is a walled garden" complaints.

dnlbtz:
This also fixes a second bug that looks unrelated: on main, satellite stops
sending the Wayland selection to X for as long as any X11 client owns
CLIPBOARD. No Electron and no focus edge case needed.

Repro on main (v0.8-12-g0dde7ca, niri 26.04), keeping the X11 window focused
throughout:

    RUST_LOG=debug xwayland-satellite :77 &
    env -u WAYLAND_DISPLAY DISPLAY=:77 alacritty &   # focus this one

    printf %s one | wl-copy
    DISPLAY=:77 xclip -o -selection clipboard        # one

    printf %s x-owner | DISPLAY=:77 xclip -selection clipboard -in &
    printf %s two | wl-copy
    DISPLAY=:77 xclip -o -selection clipboard        # x-owner, expected two

    pkill -x xclip
    printf %s three | wl-copy
    DISPLAY=:77 xclip -o -selection clipboard        # three

The debug log has "Clipboard set from Wayland" for the first and third copies
and nothing at all for the second, so set_clipboard is never reached.
new_selection in src/server/selection.rs returns None while state.source is
SelectionData::X11, which drops the offer and leaves the X selection alone.

On c512773 the second copy gives "two" and the log line is there. I found this
through an xpra client, which owns CLIPBOARD for its whole session and so hits
it permanently; on the PR build that works too, both directions.

21Rebel:
Datapoint, since #485 mentions this PR fixing things there: on niri 26.04 it does
not fix the case I ran into. I tested it twice, once as submitted and once
rebased onto current `main` (the rebase was clean, `main` has not touched
`src/server/selection.rs` or `clientside.rs` since this branch's base).

niri does advertise `ext_data_control_manager_v1`, so the fallback should be
available. Fresh instance, X11 window focused:

```
1) wayland "AAA"  -> x11 reads: 0 bytes
2) wayland "BBB"  -> x11 reads: 0 bytes
3) x11 "CCC"      -> wayland reads: BBB      (X11 -> Wayland still fails)
4) wayland "DDD"  -> x11 reads: CCC          (stuck from here on)
5) wayland "EEE"  -> x11 reads: CCC
```

The log shows the same pattern as stock 0.8.2: after `CLIPBOARD set from X11`
there is never another `Clipboard set from Wayland`, while `Primary set from
Wayland` keeps coming for the same copies.

That is probably a different path from the no-keyboard-serial case this PR was
written for, so this is not meant as a complaint about the PR. Details, a
reproducer and the code path I ended up at are in
https://github.com/Supreeeme/xwayland-satellite/issues/485#issuecomment-5542600511

PR #436, opened 2026-05-24
feat: `-enable-ei-portal` flag for XWayland emulated input support

### LINKED ISSUE #432: Handle XTEST extension
**Is your feature request related to a problem? Please describe.**
Steam input uses the XTEST extension to provide desktop input emulation. This obviously doesn't work in a wayland session.

**Describe the solution you'd like**
Satellite should hook into XTEST calls and translate them to wayland either using the xdg input emulation portal or libuinput (which is what libextest currently does).

**Describe alternatives you've considered**
Obviously libextest exists, but it's a bit cumbersome to `LD_PRELOAD` it for individual apps (especially desktop ones, where either the binary itself or the desktop item need to be modified), and 32- vs 64-bit becomes an issue.

**Additional context**
See nixos/nixpkgs#520162. I've been told that many compositors with built-in xwayland support do this already.

This might be made much simpler given that you've already put a lot of work into extest.

justDeeevin:
closes #432

i figured i would follow the pattern you've got for other flags of making them user-configurable and generally pass-through. if you'd prefer it just always being on or being opt-out, i can make that change.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -101,6 +101,7 @@ pub fn main(mut data: impl RunData) -> Option<()> {
             &xsock_xwl.as_raw_fd().to_string(),
             "-displayfd",
             &ready_tx.as_raw_fd().to_string(),
+            "-enable-ei-portal",
         ])
         .args(data.flags())
         .env("WAYLAND_DISPLAY", socket.socket_name().unwrap())


justDeeevin:
could also be worthwhile to just. allow users to specify any additional args they'd like passed to XWayland. wdyt?

Supreeeme:
I don't think there's really a scenario where this flag could be enabled but shouldn't be. I would have this always on as long as it doesn't negatively affect compositors. If you can, I would check that this doesn't negatively affect compositors that don't support this while it's enabled too.

See #335 for why we don't allow users to specify just any flags.

justDeeevin:
lol now it's a literal one-liner

justDeeevin:
@Supreeeme following up :3

Supreeeme:
Sorry for the delay. I don't really have a problem with adding this flag itself. But I just went and manually tested this on with Steam on Niri, and it did not work. Did it work for you?

justDeeevin:
I only just got around to actually testing it myself and I can confirm that it doesn't work... I have two theories
1. that flag just plain doesn't do what we're expecting it to (unlikely since I've been told this is the behavior on KDE)
2. This:  
  > This option has no effect if the compositor doesn't support the relevant XDG portal or **if Xwayland was not compiled with EI and OEFFIS support**.
  
  Is getting in the way. How could we check what compile options were used for the running xwayland?

PR #437, opened 2026-05-31
fix: handle late scale events for removed outputs

awsms:
Compositor output reconfiguration can remove an output global while events for the old `wl_output` object are still queued. During removal we strip `OutputScaleFactor` from the output entity, but a later `wl_output.scale` event still tried to update that component and unwrapped the missing query result.

Treat a missing `OutputScaleFactor` as a stale output event and ignore it instead of panicking. Add a regression test that removes an output and then delivers a late scale event for the same `wl_output`.

### DIFF
--- a/src/server/event.rs
+++ b/src/server/event.rs
@@ -1026,7 +1026,10 @@ fn update_output_scale(
     mut output_scale: hecs::QueryOne<&mut OutputScaleFactor>,
     factor: OutputScaleFactor,
 ) -> bool {
-    let output_scale = output_scale.get().unwrap();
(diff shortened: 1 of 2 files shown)


En-En-Code (approved):

OJ:
Thanks for the work on this folks! Quick question: do you have any idea when the next release is due to make an appearance? As a KVM user this one bites me every day, hence would love to see it released. I can patch this into my Nixos config but would prefer to stick to release cycles where possible.

I appreciate everyone's efforts. Cheers.

Issue #66, opened 2024-10-27
Mouse pointer does not work with some games


tyressk:
In games where there is a mouse pointer which can be used to select a menu option for example, there are some issues with the cursor  (using Niri as compositor)

For example:

Borderlands 3 - hovering over any menu items does not select it
Supermarket Together - focuses on things on the right side of the screen (language selection) even when you are focusing on the left "New Game" menu items, etc, so it becomes impossible to play the game

The cursor works ok on Gnome

Supreeeme:
Hmm, this sounds like the output positioning issues. Are you using the latest git commit? There have been quite a few fixes since 0.4.

tyressk:
Yes @Supreeeme , git log says I'm on commit 67efa2c559b53d3ca90c131c06c247278783880d . I also did cargo clean just in case. I could have sworn the Borderlands 3 menu clicking worked before a recent `git pull`, but I went through the past commits after 0.4 and couldn't make it work on any of them

Supreeeme:
Could you run satellite with RUST_LOG=1 and post the log while running those games?

tyressk:
Sure, this is the output for both games. Both start when `RUST_LOG=1 cargo run` was called, game was started from launcher, and run until the first menu (where the mouse cursor can't click some items)

Borderlands 3: https://gist.github.com/tyressk/e11013737d3cbe418b7002c657fb8b7e

For Supermarket Together:
```
2024-10-27T19:29:51.395Z INFO  xwayland_satellite > Connected to Xwayland on :0
 2024-10-27T19:29:51.404Z INFO  xwayland_process   > The XKEYBOARD keymap compiler (xkbcomp) reports:
 2024-10-27T19:29:51.404Z INFO  xwayland_process   > > Warning:          Unsupported maximum keycode 708, clipping.
 2024-10-27T19:29:51.404Z INFO  xwayland_process   > >                   X11 cannot support keycodes above 255.
 2024-10-27T19:29:51.405Z INFO  xwayland_process   > > Warning:          Could not resolve keysym XF86KbdInputAssistPrevgrou
 2024-10-27T19:29:51.405Z INFO  xwayland_process   > > Warning:          Could not resolve keysym XF86KbdInputAssistNextgrou
 2024-10-27T19:29:51.406Z INFO  xwayland_process   > Errors from xkbcomp are not fatal to the X server
[...]

PR #439, opened 2026-06-10
fix: guarantee `wl_output::geometry` is sent on global bind 

En-En-Code:
Although this could manifest as the "early mouse border" bug, this is not a fix to #66 it was introduced in 33c344fe, which that issue is older than.

### DIFF
--- a/testwl/src/lib.rs
+++ b/testwl/src/lib.rs
@@ -581,10 +581,23 @@ impl Server {
     /// This function must be called after the globals have been dispatched in order to use the
     /// output on the server side created by `new_output` (this function's return value).
     #[track_caller]
-    pub fn finalize_output(&mut self) -> WlOutput {
+    pub fn finalize_output(&mut self, x: i32, y: i32) -> WlOutput {
         let output_s = self.state.last_output.take().expect("No new outputs");
         let output_data = self.state.outputs.get_mut(&output_s).unwrap();
         output_data.global_id = self.state.last_output_global.take();
+        output_s.geometry(
+            x,
+            y,
+            0,
+            0,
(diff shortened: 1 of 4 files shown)


En-En-Code:
Okay, so I guess something I did broke `forced_1x_scale_consistent_x11_size`, but only on stable release. Works on bleeding edge Xwayland. What the... I'll figure this out later.

Edit: Solved. Forget I also needed to set the initial Xdg output events again. And of course the CI is deadlocked on a stupid clipboard test again.

testwl/src/lib.rs
```diff
@@ -581,10 +581,23 @@ impl Server {
...
+            0,
+            0,
+            wl_output::Subpixel::None,
+            "xwls".to_string(),
+            "fake monitor".to_string(),
+            wl_output::Transform::Normal,
+        );
+        output_s.mode(wl_output::Mode::Current, 1000, 1000, 0);
+        output_s.done();
+        self.dispatch();
```
Supreeeme, line 600: Why did this move?

En-En-Code: I needed it to. The `wl_output::geometry` events and the `xdg_output::logical_*` events need to be dispatched from the event queue within the same dispatch, since updating the global output offset while `OutputDimensionsSource` is still `Wl` sends a `wl_output::geometry` event to Xwayland, which is the very event which dispels the bug.

The original `GlobalDispatch` sent these events before the associated `xdg_output` was created, too early to work with easily. By pushing it to `finalize_output`, tests that just need a basic monitor initialized pass unchanged, while tests which need fine-grained control over the output events sent have it.

Issue #209, opened 2025-07-29
RRScreenChangeNotify is sent on every focus change


corngood:
### Operating System

nixos (linux)

### Wayland Compositor

sway

### Describe the issue.

[As discussed here](https://github.com/Supreeeme/xwayland-satellite/issues/80#issuecomment-3120655893)

Changing the primary display sends `RRScreenChangeNotify`.  xfreerdp listens for that event and does some heavy resizing work, flickers, and picks up the resolution from the display where the mouse is.  This happens whenever I move focus between X11 apps on different outputs since 03a53b6ad7bd3d4b379f660d813ac75731d68641.  Part of it is bad behaviour in the app, but I think it's reasonable for the app to assume the screen layout isn't changing as frequently as focus.

I think it would be a good idea to make this behaviour configurable, and perhaps disable it by default.  Some compositors do appear to support the idea of a primary output:

https://wayland.app/protocols/kde-primary-output-v1

So it could be something like `--primary-output=auto|focused|none`.  I can start a PR for this if you think it's a good idea.

> It seems sattelite chooses its primary monitor differently to the other xwayland implementations I've tested so far.

There's still something going on here that I don't understand.  If sway doesn't set a primary, and xwayland-satellite didn't either (when this was tested), what explains the behaviour difference?  Just the order of the outputs?

### Please confirm you have reproduced this on the latest commit.

- [x] I have reproduced this on the latest commit

corngood:
> Before adding a command line option for it, I'd first want to see if it would be enough to just start off with the current automatic primary display behavior, and then stop it once satellite gets some xrandr event from xrandr --primary that it didn't trigger, and re-enable it after xrandr --noprimary.

@Supreeeme 

In general, shouldn't we be aiming to match behaviour to the underlying compositor?  I've only tested sway, but it doesn't seem to set a primary at all.  Are there any compositors that do this focus-following behaviour?

Supreeeme:
Compositors don't set a primary display because Wayland provides no universal mechanism for signaling a primary display, yet games rely on this behavior as seen in the previous issue. So satellite needs to do something here.

corngood:
> Compositors don't set a primary display because Wayland provides no universal mechanism for signaling a primary display

If it's a compositor working as an X11 server, then it doesn't matter if wayland has a mechanism.  Sway could set it on Xwayland using RANDR just like satellite does.

> So satellite needs to do something here.

[...]

Issue #80, opened 2024-12-06
Only vertical/incorrect resolutions suggested in apps


JimmJam:
Across some poorly coded applications which don't allow resolutions outside of what is initially seen by the checking what the output supports, notably some older games, only strange resolutions like 800x1280 are suggested by default.

I don't believe this to be a "primary monitor" issue, despite having a vertical display, as changing which ports the displays were plugged into did not seem to change anything, nor did changing from Niri to Sway (still using xwayland-sattelite).

![Screenshot from 2024-12-06 15-27-27](https://github.com/user-attachments/assets/26fa3e50-213d-4508-8dfb-e47313439f4e)
![Screenshot from 2024-12-06 15-18-44](https://github.com/user-attachments/assets/ec49b61e-c1a3-458a-aca9-77c963490c4c)

Supreeeme:
What's your monitor setup?

JimmJam:
In order of direction, HDMI-A-1 at 1080p, DP-1 at 1080p, DP-2 at 1080p, and DP-3 at 1680x1050 rotated 90 degrees. The last seems to be the monitor Satellite is locking onto for its assumed resolution.

Supreeeme:
What games are in those screenshots? And could you post a satellite log (with `RUST_LOG=debug`) of you just starting up one of these games?

JimmJam:
Sure thing! The games in question that were being ran were Dishonoured and Far Cry 3, here's logs of me starting sattelite, then Steam, then the game:

Dishonoured: [dishonoured.log](https://github.com/user-attachments/files/18130064/dishonoured.txt)
Far Cry 3: [farcry3.log](https://github.com/user-attachments/files/18130097/farcry3.txt)
Hitman 3: [hitman3.txt](https://github.com/user-attachments/files/18130836/hitman3.txt)

Also, I should clarify now that I've looked through the logs, the layout of my displays is as output by `niri msg outputs | grep -E 'Logical|Scale|Transform'`

```
  Logical position: 5760, 0
  Logical size: 1050x1680
  Scale: 1
  Transform: 90° counter-clockwise
  Logical position: 3840, 405
  Logical size: 1920x1080
  Scale: 1
  Transform: normal
  Logical position: 1920, 405
  Logical size: 1920x1080
  Scale: 1
  Transform: normal
  Logical position: 0, 405
  Logical size: 1920x1080
  Scale: 1
  Transform: normal
```

Supreeeme:
[...]

Issue #429, opened 2026-05-06
Implement X11 overlay (`override_redirect`)


bczhc:
### Operating System

Arch Linux

### Wayland Compositor

niri

### Describe the issue.

I tried [`activate-linux`](https://github.com/MrGlockenspiel/activate-linux) with xwayland-satellite, and it's not working properly. On X, it starts as an overlay at the screen bottom-right corner.

There's a simple C program displaying a similar overlay at the bottom-right to demonstrate this (_small disclaimer_: LLM-generated because i'm not quite familiar with x11):

<details>
<summary>code</summary>

```c
// Compile using: cc `pkg-config --cflags x11 xfixes cairo` `pkg-config --libs x11 xfixes cairo` main.c 

#include <X11/Xlib.h>
#include <X11/Xutil.h>
#include <X11/Xatom.h>
#include <X11/extensions/Xfixes.h>
#include <X11/extensions/shape.h>
#include <cairo/cairo.h>
#include <cairo/cairo-xlib.h>
#include <signal.h>
#include <stdio.h>
#include <stdlib.h>
#include <unistd.h>

#define OVERLAY_W 260
#define OVERLAY_H 80
#define MARGIN     30

static volatile sig_atomic_t running = 1;

static void sig_handler(int sig) {
    (void)sig;
    running = 0;
}

static void draw(cairo_t *cr, cairo_surface_t *surface, Display *dpy) {
    cairo_save(cr);

    /* transparent background */
    cairo_set_source_rgba(cr, 0, 0, 0, 0);
    cairo_set_operator(cr, CAIRO_OPERATOR_SOURCE);
    cairo_paint(cr);
    cairo_restore(cr);

    /* rounded background pill */
    double x = 10, y = 10, w = OVERLAY_W - 20, h = OVERLAY_H - 20, r = 12;
    cairo_new_sub_path(cr);
    cairo_arc(cr, x + w - r, y + r,     r, -0.5 * 3.14159, 0);
    cairo_arc(cr, x + w - r, y + h - r, r, 0,              0.5 * 3.14159);
    cairo_arc(cr, x + r,     y + h - r, r, 0.5 * 3.14159,  1.0 * 3.14159);
    cairo_arc(cr, x + r,     y + r,     r, 1.0 * 3.14159,  1.5 * 3.14159);
    cairo_close_path(cr);
    cairo_set_source_rgba(cr, 0.08, 0.08, 0.08, 0.65);
    cairo_fill(cr);

[...]

PR #442, opened 2026-06-13
fix: don't override RANDR primary output on focus change

zakissimo:
## Summary

`focus_window()` called RANDR `SetOutputPrimary` on every cross-output focus change (the "primary follows focus" behavior from #80). Each call emits `RRScreenChangeNotify`, and under focus-follows-mouse with windows on multiple monitors it fires on every pointer hover — so apps that rebuild or hide their windows on that event (xfreerdp, Sierra Chart under Wine) flicker or vanish. Tracked as #209.

This PR stops satellite from driving the primary at all. The primary is now the user's to set: `xrandr --primary <output>` sticks instead of being overwritten on the next focus change, matching a bare Xorg/sway session.

## Behavior change

- Primary no longer follows focus — no more `RRScreenChangeNotify` churn (#209).
- An explicit `xrandr --primary <output>` persists.
- The automatic behavior from #80 is dropped; those users set the primary once (the #80 reporter already confirmed `xrandr --output X --primary` fixes their game).

## Notes

Per the discussion this replaces the earlier env-var opt-out with simply removing the focus-following, which @Supreeeme was open to testing. I kept the RANDR output tracking (`update_outputs`, the output map) in place, so a later "set once at startup" or "defer to xrandr" refinement can build on it without re-adding anything.

The `primary_output` integration test is rewritten to assert the new contract: focus does not set the primary, and an explicitly set primary is not overridden by a subsequent focus.

### DIFF
--- a/src/server/event.rs
+++ b/src/server/event.rs
@@ -223,11 +223,6 @@ impl SurfaceEvents {
                         },
                         connection,
                     );
-                    if state.last_focused_toplevel == Some(*window) {
-                        let output = get_output_name(Some(&on_output), &state.world);
-                        debug!("focused window changed outputs - resetting primary output");
-                        connection.focus_window(*window, output);
-                    }
 
                     if state.fractional_scale.is_none() {
                         let output_scale = output_data.get::<&OutputScaleFactor>().unwrap().get();
--- a/src/xstate/mod.rs
+++ b/src/xstate/mod.rs
@@ -1437,35 +1437,6 @@ impl XConnection for RealConnection {
         }) {
             debug!("ChangeProperty failed ({window:?}: {e:?})");
         }
-
-        if let Some(name) = output_name {
-            let Some(output) = self.outputs.get(&name).copied() else {
-                warn!("Couldn't find output {name}, primary output will be wrong");
-                return;
-            };
-            if output == self.primary_output {
-                debug!("primary output is already {name}");
-                return;
-            }
-
-            if let Err(e) = self
-                .connection
-                .send_and_check_request(&xcb::randr::SetOutputPrimary { window, output })
-            {
-                warn!("Couldn't set output {name} as primary: {e:?}");
-            } else {
-                debug!("set {name} as primary output");
-                self.primary_output = output;
-            }
-        } else {
-            let _ = self
-                .connection
-                .send_and_check_request(&xcb::randr::SetOutputPrimary {
-                    window,
-                    output: Xid::none(),
-                });
-            self.primary_output = Xid::none();
-        }
     }
 
     fn close_window(&mut self, window: x::Window) {
--- a/tests/integration.rs
+++ b/tests/integration.rs
@@ -1535,30 +1535,41 @@ fn primary_output() {
     assert_ne!(surface1, surface2);
     f.wait_and_dispatch();
 
+    // satellite must no longer drive the RANDR primary output from focus (#209):
+    // focusing windows, including across outputs, must leave the primary untouched.
     f.testwl.focus_toplevel(surface1);
     std::thread::sleep(std::time::Duration::from_millis(10));
     let reply = conn.get_reply(&xcb::randr::GetOutputPrimary { window: conn.root });
-    assert_eq!(reply.output(), output1);
+    assert_eq!(
+        reply.output(),
+        Xid::none(),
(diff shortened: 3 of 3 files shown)


En-En-Code:
Thanks for going through the prior art. I was in this code recently tracking an unrelated bug and was confused about this primary output switching behavior. Having context is illuminating.

I believe the behavior is just wrong, as I expect programs using `override_redirect` on `root` (see #429) attempt to choose the primary output to display to. Notifications and dock bars (if using `wl_layer_shell` proves to work for the former, it should also for these) I expect even moreso, and it is more obviously incorrect if your status bar repeatedly jumps monitors. I say "expect" having not read any of the relevant X programs.

The "minimal CLI options," as I understand it, means a "minimal user-facing complexity," which a new environment variable and CLI flag add equally. That was written before `satellite` started forwarding more Xserver/Xwayland flags and starting documenting them in its `man` page. Still, adding a new option is a version bump due to the ability of satellite integrations check the version to determine which flags it can use.

zakissimo:
Totally agree. Debugging this, the only reason `xrandr --primary` doesn't work today is that satellite overwrites it on the next focus change — so the fix and the "sensible solution" for #80 turn out to be the same thing: stop driving the primary programmatically and let the user own it, like a normal X server.

Confirmed on my setup with the per-focus reset disabled: `xrandr --output DP-3 --primary` sticks and satellite never touches it again; with no user command there's simply no primary set — same as bare sway, the baseline corngood described.

Proposal:
- Drop the per-focus `SetOutputPrimary`. That ends the whole class at once — Wine/RDP rebuilds, and status bars / docks / notifications / the #429 overlays jumping monitors — because the primary stops moving.
- #80 users run `xrandr --primary <output>` once (or from their compositor/login config) and it persists. The #80 reporter already confirmed `xrandr --output X --primary` fixes their game, so this just makes it stick instead of fighting it.

Net: respect the user's explicit choice instead of guessing from focus. The only regression is #80 goes from automatic to "set it once" — the conventional Xorg/sway model anyway.

If that sounds right to you and @Supreeeme, I'll rework this PR to remove the focus-following rather than gate it behind a flag. (If an automatic default is still wanted for #80, picking the primary once at startup would at least avoid the churn — but I'd lean toward just respecting `xrandr`.)

En-En-Code:
I do not see `satellite` setting primary to a best-guess at start-up to be incompatible with deferring to `xrandr` commands. I even was thinking of a more powerful solution--having a `--primary` flag which instead of taking a behavior (`auto|follow|...`) take an output and set that to be primary. It could also be a seperated list of outputs, picking the first one on the list which exists. This would allow for the initial primary output to be removed (and reconnected) and the primary would shift automatically.

Considering the breadth of design space available, I also want @Supreeeme 's opinion before more code on this matter is written.

Supreeeme:
Like I mentioned in #80 I'd prefer to avoid command line arguments. I'd be fine with satellite not adjusting the primary display if it gets set by xrandr. I don't remember if there would be ramifications to not switching the primary display when output focus changes but I'd be open to testing it out.

zakissimo:
Went with simply not touching the primary rather than the env var — pushed, ready to test.

On "don't adjust if it gets set by xrandr": I tried that first and it's racy under focus-follows-mouse. Satellite re-set the primary on every pointer hover, so it overwrote a manual `xrandr --primary` on the very next hover, and `RRScreenChangeNotify` carries neither the originating client nor the new value — so satellite can't reliably distinguish "the user set this" from "I set this" before clobbering it again. Not driving the primary at all sidesteps the detection problem entirely: `xrandr --primary <output>` just sticks.

The only regression is #80 going from automatic to set-once, which its reporter already confirmed works (`xrandr --output X --primary` fixed their game). I left the output tracking in place, so if you'd rather set a primary once at startup instead, that can build on top without re-adding anything.

En-En-Code:
I still have a couple of concerns which this design. The most important is that disconnecting and reconnecting a monitor will not restore that monitor to primary status. Needing an `xrandr` command every time you close and reopen your laptop lid is certainly not good user experience. I also think setting primary once on initialization would be nice. [SDL3 has heuristics for which output to choose](https://github.com/libsdl-org/SDL/blob/main/src/video/wayland/SDL_waylandvideo.c?plain=1#L246-L319), although since the user can still change the primary afterwards, we may not need something so complex.

PR #447, opened 2026-06-27
fix: allow unsetting min/max size hints after a WM_NORMAL_HINTS PropertyNotify

jturcotte:
The latest version of Webex would have a locked size window because of this issue, while it worked as expected in Gnome and Sway.

The xdg_toplevel protocol for set_min_size and set_max_size states: "As a result, a client wishing to reset the maximum size to an unspecified state can use zero for width and height in the request."


En-En-Code (approved):
Cool. Can't say I ever expected a program to unset `PMinSize` and `PMaxSize` from `WM_NORMAL_HINTS`. Makes sense Wayland also has a reset mechanism. Guess I must have never read that piece of the spec.

Supreeeme:
Wow, I use Webex every day and just assumed the client was wack :smiley:

PR #452, opened 2026-07-07
feat(scaling): add compositor-side supersampling, EWMH maximization, and input region scaling

### LINKED ISSUE #180: Mixed DPI behaviour with Niri not taking into account scaling factor
### Operating System

Manjaro testing

### Wayland Compositor

Niri

### The Problem

New behaviour of xwayland-satellite related to scaling, namely «Satellite will now always report native monitor resolutions to Xwayland» and «Note that for mixed DPI monitor setups, satellite will choose the smallest monitor's DPI, meaning apps may have small text on other monitors.» seems to be very inconvenient with mixed DPI setups.

This is related to two monitor Niri setup like this:

output "Philips Consumer Electronics Company PHL 278E1 XXX" {
    mode "3840x2160"
    scale 1.5
    position x=0 y=0
    focus-at-startup
}

output "Philips Consumer Electronics Company PHL 242E2F YYY" {
    mode "1920x1080"
    scale 1.0
    position x=2560 y=180
}

And to the situation when external monitor is connected to a notebook with HiDPI display _before_ the start of Niri itself.

Xwayland-satellite seems to have no key to restore pre-0.6-release behaviour which seemed to take into account configured output scaling factors.

### System Information

* niri version:  niri 25.05.1 (8ba57fc)
* xwayland-satellite version: xwayland-satellite version v0.6-3-gda2ecb5

### Steps to reproduce

1. Install the latest Niri and Xwayland-satellite releases or the latest commits.
2. Connect two monitors and configure it with different effective DPI.
3. Start Niri and run any X application in Niri session configured with Xwayland-satellite as an X server backend runner.

### LINKED ISSUE #392: Forza Horizon 5 starts "embedded" into steam

### LINKED ISSUE #449: Steam windows interactable area is cut off
### Operating System

CachyOS

### Wayland Compositor

Niri

### Describe the issue.

I use a laptop with an external widescreen monitor, the laptop lid is closed and so its screen is off.
So only the widescreen monitor is active

When I open steam, only a portion of the steam window is interactable (red area on the image)

Image

It only happens with steam but it's the only X11 app I use

### Steps to reproduce

Open steam on an external monitor with a different image ratio, probably ?

### LINKED ISSUE #456: Maximized X11 window starts transparent with incorrect input region on scale-2 output
### Operating System

Gentoo

### Wayland Compositor

Niri

### Describe the issue.

An Avalonia-based X11 application can start partially or fully transparent when it restores a persisted maximized window state on a scale-2 output.

The visible contents, X11 window geometry, and pointer input region appear to become desynchronized. Parts of the window may be transparent, mouse clicks do not reach the expected controls, and toggling fullscreen/maximized state does not repair it.

The same application and scaling configuration work correctly under KDE Plasma/KWin.

The issue does not occur if the application starts with a normal window state. After the window has been mapped and rendered normally, maximizing it manually works correctly.

Resetting the persisted window state from `Maximized` to `Normal` fixes the next launch.

This looks similar to #449 because both issues involve the visible window and interactable region having different dimensions.

### Environment

- niri: `26.04 (8ed0da4)`
- xwayland-satellite: `0.8.1`
- Xwayland: `24.1.13`
- GPU: NVIDIA GeForce RTX 4060 Ti
- NVIDIA driver: `595.84`
- Monitor mode: `3840x2160 @ 59.997 Hz`
- Logical output size: `1920x1080`
- Output scale: `2`
- Application: SourceGit `2026.15`
- UI framework: Avalonia `11.3.18`
- Application scaling: `AVALONIA_GLOBAL_SCALE_FACTOR=2`

The problem was also reproduced with SourceGit `2026.14`, so it does not appear to be a regression specific to SourceGit 2026.15.

### Steps to reproduce

1. Start a niri session with xwayland-satellite.
2. Configure the output with scale `2`.
3. Launch SourceGit with:

   ```sh
   AVALONIA_GLOBAL_SCALE_FACTOR=2 /opt/sourcegit/sourcegit
4. Maximize the SourceGit window.
5. Close SourceGit while it is maximized. SourceGit persists the state as:
"LauncherWindowState": 2
 6. Launch SourceGit again with the same environment.
 7. Observe that the window is partially or fully transparent.
 8. Observe that pointer input does not match the visible controls and fullscreen/maximize controls do not work correctly.
 9. Stop SourceGit and change the saved state to:
"LauncherWindowState": 0
10. Launch SourceGit again.
11. The window now renders correctly and remains usable.
12. Maximizing the already-rendered window manually also works correctly.

Expected behavior
Restoring a maximized X11 window should produce a fully rendered window whose input region matches its visible geometry.
Restoring the maximized state at startup should behave the same as maximizing the window manually after it has been mapped.
Actual behavior
When maximized state is restored during startup:
- the window is partially or fully transparent
- the rendered area may cover only part of the allocated window
- pointer input does not match the visible content
- fullscreen/maximize operations appear ineffective

Starting in normal state and maximizing afterward works correctly.
Control test
The same SourceGit build, AVALONIA_GLOBAL_SCALE_FACTOR=2, monitor, GPU, and persisted maximized state work correctly under KDE Plasma/KWin.
Relevant application code
SourceGit restores the persisted state from Window.OnOpened():
protected override void OnOpened(EventArgs e)
{
    base.OnOpened(e);

    var preferences = ViewModels.Preferences.Instance;
    WindowState = preferences.Layout.LauncherWindowState;
}
Source:
https://github.com/sourcegit-scm/sourcegit/blob/v2026.15/src/Views/Launcher.axaml.cs
Logs
There are no GPU or Xwayland errors in the kernel journal when the issue occurs.
The only SourceGit messages are unrelated session-management and input-method warnings:

[X11Platform] SMLib/ICELib reported a new error:
SESSION_MANAGER environment variable not defined

[FreeDesktopPlatform] GetNameOwnerAsync failed:
org.freedesktop.DBus.Error.NameHasNoOwner:
Could not get owner of name 'org.fcitx.Fcitx'
Workaround
Force the application to start with a normal window state, then maximize it manually after it has rendered.

### LINKED ISSUE #463: Override-redirect popups / positioned windows are mispositioned on scaled outputs (not Steam-specific)
### Operating System

Arch Linux

### Wayland Compositor

Niri

### Describe the issue.

On an output with scale ≠ 1, XWayland windows that position themselves (override-redirect popups, self-positioned floating windows) land in the wrong place. The error is absent at scale 1 and appears at scale 2, so it looks like a missing scale conversion rather than an app bug.

This is not Steam-specific: it also affects the fcitx5 candidate window, which is a separate X11 override-redirect window created by fcitx, not by Steam.

Environment
xwayland-satellite: <version/commit>, niri <version>, OS <...>
Output: native <e.g. 3840x2160>, tested at scale 2 and scale 1
Reproduce
niri msg output <name> scale 2
Launch an XWayland app that creates positioned popups / floating windows.
Observe the popup/window location; then scale 1 and compare.
Observed

At scale 2: Steam context menus / notifications / floating windows are offset; the fcitx5 candidate window (typing into Steam) appears upper-right instead of under the cursor. At scale 1: all placed correctly.

<optional but valuable: pixel offset at scale 1 vs scale 2; and one non-Steam X11 app whose menus are offset, e.g. xterm Ctrl+click menu>

Likely cause

Per ARCHITECTURE.md, override-redirect popups have their requested position respected. Since v0.6 (#180) satellite reports native resolution, so that position is in physical pixels. If it's placed into the compositor's logical space without dividing by the output scale, the popup lands scale× too far — matching the behavior (0 error at scale 1, offset at scale 2). Same "coordinate not scaled by surface scale" class as the tablet-input bug fixed in #179 (and #172 for touch); #179's fix is a likely template.

Related
#156 (Steam menu mislocation) attributes it to Steam restoring its saved window position. This report is the scale-dependent offset, which is independent of that and also hits the non-Steam fcitx candidate — possibly distinct or compounding.
#301 is the fcitx size counterpart; this is the position one.
Expected

Positioned popups / floating windows should land at the same on-screen location at scale 2 as at scale 1 (respected X11 coordinate scale-converted into logical space).

### Steps to reproduce

_No response_

gimgiwer:
### What

Adds optional compositor-side supersampling for mixed-DPI setups, implements missing EWMH window maximization, and fixes Wayland input region coordinate scaling for viewport-managed surfaces.

* **Compositor supersampling**: Renders X11 apps at the maximum output DPI and lets the Wayland compositor downscale them on lower-DPI monitors via `wp_viewport`. 2D monitor layout is resolved through a minimum spanning tree. Includes `XWAYLAND_SATELLITE_BASE_SCALE` (1.0–8.0) to override the scale on low-end systems.
* **Input region scaling**: Buffers and scales `wl_region` rectangles to match destination logical coordinates, eliminating click offsets and destroying replaced `wl_region` objects immediately.
* **EWMH maximization**: Adds `_NET_WM_STATE_MAXIMIZED_VERT` and `_NET_WM_STATE_MAXIMIZED_HORZ` with pre-mapping state caching so frameworks like Avalonia (SourceGit) map directly in the maximized state.
* **Non-blocking selection**: Converts clipboard transfers to `O_NONBLOCK` pipes with a 2-second timeout and a 16 MiB payload cap to prevent X11 event loop stalls.

### Why
On mixed-DPI setups (for example, a 2560x1600 1.75x internal laptop display paired with a 1080p 1.0x external monitor),  Xwayland forces a single global scale across outputs. Without compositor scaling, X11 apps like Steam end up blurry on the HiDPI screen or tiny on standard screens, with geometry gaps when crossing monitor boundaries.

Additionally:
* Apps like SourceGit start unmaximized because `_NET_WM_STATE` maximization atoms were unhandled.
* Clicks inside scaled windows registered at the wrong offsets because `wp_viewport` requires logical destination coordinates, while X11 passes buffer pixels.
* Slow or broken clipboard transfers could block the main poll loop.

### How

* **Compositor Scaling & Spanning Tree**:
  * Gated by `XWAYLAND_SATELLITE_COMPOSITOR_SCALING` (checked for non-empty, non-zero values).
  * Replaced 1D coordinate offsets with a 2D minimum spanning tree solver (Prim's algorithm in `recalculate_x11_output_positions`) rooted at the monitor closest to `(0, 0)`.
  * Delta offsets between adjacent monitors are scaled by the boundary edge scale factor (`dx * scale_x * global_scale`) to prevent gaps or overlaps on mixed-DPI and rotated setups.
  * Added `XWAYLAND_SATELLITE_BASE_SCALE` (clamped to `[1.0, 8.0]`): setting it to `1.0` keeps native resolution and saves VRAM without breaking multi-monitor positioning fixes.
  * Hardened MST offset lookups with `unwrap_or(0)` against incomplete output graphs during monitor hotplug.
  * Titlebar heights are dynamically calculated from font line metrics (`ascent - descent + line_gap + padding`) and cached via `OnceLock<i32>`.
* **Input Region Scaling**:
  * While X11 passes `SetInputRegion` rectangles in buffer pixels, `wp_viewport` expects logical destination coordinates.
  * Operations (`Add`/`Subtract`) are buffered in `RegionData.ops` (`Arc<Mutex<Vec<RegionOp>>>`) and scaled by `(scale_x, scale_y)` before issuing Wayland requests.
  * Stale `wl_region` objects are destroyed immediately on replacement (`old.destroy()`) and on surface destruction to avoid object leaks.
  * Applied aspect ratio correction for 90°-rotated fullscreen surfaces.
  * In `WlTouch::Up` and `WlTouch::Cancel`, `CurrentSurface` is cleanly removed to avoid stale touch tracking.
* **EWMH Maximization**:
  * Advertised `_NET_WM_STATE_MAXIMIZED_VERT` and `_NET_WM_STATE_MAXIMIZED_HORZ` in `_NET_SUPPORTED`.
  * `update_net_wm_state()` performs atomic read-modify-write on `_NET_WM_STATE` properties without clobbering existing flags like fullscreen.
  * Pre-mapping state buffering in `WindowAttributes` ensures windows that set state before `XMapWindow` (Avalonia/SourceGit) start already maximized.
  * Synced bidirectionally with `xdg_toplevel::Event::Configure`.
* **Clipboard & Selection Hardening**:
  * Switched pipe I/O to `O_NONBLOCK` with an explicit 2-second timeout (`PendingTransfer`).
  * Capped payload size at 16 MiB (`MAX_CLIPBOARD_SIZE`), returning `ATOM_NONE` on overflow to prevent memory exhaustion and hanging clients.
  * The main loop polls with stack-allocated `[PollFd; 5]` when idle, only heap-allocating `Vec<PollFd>` during active transfers.

### Related Issues

Closes #449
Closes #456
Related #180

### Testing

Tested on Arch Linux with Niri compositor on AMD Ryzen 7 780M (1.75x 2560x1600) + external 1080p (1.0x):
* `cargo test --all-targets` passes.
* Unit tests cover multi-scale MST layout, 90° rotated screens, monitor disconnect fallback, rapid maximization toggles, `BASE_SCALE` override, and scaled input region clipping.
* Verified sharp Steam rendering without upscale blur; window dragging across mixed-DPI screens transitions without visual gaps or jumping.
* Mouse and touch clicks register at exact pixel locations across all scale factors.
* Confirmed SourceGit starts maximized immediately on launch.

original_horizontal
patched_horizontal

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 /target
 /result
+*.patch
--- a/README.md
+++ b/README.md
@@ -45,11 +45,17 @@ which is likely after your compositor is started if it supports systemd.
 Note that you *do not* need the systemd service if the compositor already integrates xwayland-satellite (see below).
 
 ## Scaling/HiDPI
-For most GTK and Qt apps, xwayland-satellite should automatically scale them properly. Note that for mixed DPI monitor setups, satellite will choose
-the smallest monitor's DPI, meaning apps may have small text on other monitors.
+For most GTK and Qt apps, xwayland-satellite should automatically pick up the right scale.
 
-Other miscellaneous apps (such as Wine apps) may have small text on HiDPI displays. It is application dependent on getting apps to scale properly with satellite,
-so you will have to figure out what app specific config needs to be set. See [the Arch Wiki on HiDPI](https://wiki.archlinux.org/title/HiDPI) for a good place start.
+By default, satellite renders X11 apps at the highest active output scale so text stays sharp on HiDPI screens (compositor handles downscaling for lower-DPI outputs).
+
+### `XWAYLAND_SATELLITE_BASE_SCALE`
+
+Pins the render scale to a fixed multiplier (1.0–8.0) instead of picking `max(output_scale)`. Compositor-side scaling stays on, so popup positioning and coordinate mapping still line up.
+
+Set it to `1` to render at native resolution but keep correct window geometry on HiDPI displays — handy when supersampling costs too much on weak hardware.
+
+Other apps, such as Wine apps, may still need their own HiDPI settings. See [the Arch Wiki on HiDPI](https://wiki.archlinux.org/title/HiDPI) for a good starting point.
 
 Satellite acts as an Xsettings manager for setting scaling related settings, but will get out of the way of other Xsettings managers.
 To manually set these settings, try [xsettingsd](https://codeberg.org/derat/xsettingsd) or another Xsettings manager.
--- a/README.md
+++ b/README.md
@@ -83,7 +83,12 @@ Some Java applications may present themselves as a blank screen by default with 
 
 ### Scaling/HiDPI
 For most GTK and Qt apps, xwayland-satellite should automatically scale them properly. Note that for mixed DPI monitor setups, satellite will choose
-the smallest monitor's DPI, meaning apps may have small text on other monitors.
+the smallest monitor's DPI by default, meaning apps may have small text on other monitors.
+
+Alternatively, you can enable compositor-side scaling by setting the `XWAYLAND_SATELLITE_COMPOSITOR_SCALING` environment variable.
+In this mode, satellite will render X11 applications at the maximum monitor's DPI (HiDPI) and let the Wayland compositor downscale them on lower-DPI monitors.
+This ensures that apps do not look tiny or blurry on HiDPI displays in multi-monitor setups.
+You can optionally set `XWAYLAND_SATELLITE_BASE_SCALE` (1.0–8.0) to override the base scale. Setting it to 1.0 keeps native res and saves VRAM without breaking multi-monitor fixes.
 
 Other miscellaneous apps (such as Wine apps) may have small text on HiDPI displays. It is application dependent on getting apps to scale properly with satellite,
 so you will have to figure out what app specific config needs to be set. See [the Arch Wiki on HiDPI](https://wiki.archlinux.org/title/HiDPI) for a good place start.
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -39,6 +39,9 @@ pub trait RunData {
     fn flags(&self) -> &[String] {
         &[]
     }
+    fn compositor_scaling(&self) -> bool {
+        false
+    }
     fn server(&self) -> Option<UnixStream> {
         None
     }
@@ -155,7 +158,8 @@ pub fn main(mut data: impl RunData) -> Option<()> {
         }
     };
 
-    let mut server_state = EarlyServerState::new(dh, data.server(), connection);
+    let mut server_state =
+        EarlyServerState::new(dh, data.server(), connection, data.compositor_scaling());
     server_state.run();
 
     // Remove the lifetimes on our fds to avoid borrowing issues, since we know they will exist for
@@ -243,17 +247,45 @@ pub fn main(mut data: impl RunData) -> Option<()> {
             xstate.update_global_scale(scale);
         }
 
-        match poll(&mut fds, None) {
-            Ok(_) => {
-                if !fds[3].revents().is_empty() {
-                    let status = xwayland_exit_code(&mut quit_rx);
-                    if status != ExitStatus::default() {
-                        error!("Xwayland exited early with {status}");
-                    }
-                    return None;
-                }
+        let timeout = xstate.poll_timeout();
+        let is_quit = if !xstate.has_pending_transfers() {
+            let mut poll_fds = [
+                PollFd::from_borrowed_fd(server_fd, PollFlags::IN),
+                PollFd::new(&xsock_wl, PollFlags::IN),
+                PollFd::from_borrowed_fd(display_fd, PollFlags::IN),
+                PollFd::new(&quit_rx, PollFlags::IN),
+                PollFd::new(&ready_rx, PollFlags::IN),
+            ];
+            match poll(&mut poll_fds, timeout.as_ref()) {
+                Ok(_) => !poll_fds[3].revents().is_empty(),
+                Err(other) => panic!("Poll failed: {other:?}"),
             }
-            Err(other) => panic!("Poll failed: {other:?}"),
+        } else {
+            let mut poll_fds = vec![
+                PollFd::from_borrowed_fd(server_fd, PollFlags::IN),
+                PollFd::new(&xsock_wl, PollFlags::IN),
+                PollFd::from_borrowed_fd(display_fd, PollFlags::IN),
+                PollFd::new(&quit_rx, PollFlags::IN),
+                PollFd::new(&ready_rx, PollFlags::IN),
+            ];
+            xstate.collect_poll_fds(&mut poll_fds);
+            match poll(&mut poll_fds, timeout.as_ref()) {
+                Ok(_) => {
+                    let is_quit = !poll_fds[3].revents().is_empty();
+                    let events: Vec<_> = poll_fds.iter().map(|f| f.revents()).collect();
+                    xstate.process_pending_clipboard_transfers(&events);
+                    is_quit
+                }
+                Err(other) => panic!("Poll failed: {other:?}"),
+            }
+        };
+
+        if is_quit {
+            let status = xwayland_exit_code(&mut quit_rx);
+            if status != ExitStatus::default() {
+                error!("Xwayland exited early with {status}");
+            }
+            return None;
         }
     }
 }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -39,6 +39,17 @@ pub trait RunData {
     fn flags(&self) -> &[String] {
         &[]
     }
+    /// Render at the highest active output scale and let the compositor downscale.
+    /// On by default in the binary; library users may want it off.
+    fn compositor_scaling(&self) -> bool {
+        false
+    }
+    /// Pins the X11 render scale instead of picking max(output_scale).
+    /// Saves VRAM on low-end hardware without breaking HiDPI window placement.
+    /// compositor_scaling() keeps working when this is set.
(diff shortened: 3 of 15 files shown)


Supreeeme:
Haven't dug deep into the code yet, but:
1. Why is there an env var and a command line flag?
2. This behavior shouldn't be gated behind a flag, it should be default behavior. I am suspicious of this actually working properly across all applications though, I'd like to see it tested with other applications besides Steam.
3. There are some comments in Russian(?), please remove them if necessary or translate them to English.
4. Please keep the LLM yapping to a minimum, I don't enjoy reading word salad.

gimgiwer:
> Haven't dug deep into the code yet, but:
> 
>     1. Why is there an env var and a command line flag?
> 
>     2. This behavior shouldn't be gated behind a flag, it should be default behavior. I am suspicious of this actually working properly across all applications though, I'd like to see it tested with other applications besides Steam.
> 
>     3. There are some comments in Russian(?), please remove them if necessary or translate them to English.
> 
>     4. Please keep the LLM yapping to a minimum, I don't enjoy reading word salad.

1. I removed the CLI flag. Compositor-side scaling is now enabled by default, and I kept only XWAYLAND_SATELLITE_DISABLE_COMPOSITOR_SCALING as an escape hatch in case it causes regressions on some setups.
2. I agree that this should be the default behavior. My only concern is that it may still have issues across different X11 applications and mixed-DPI setups, especially because Xwayland has to render at the maximum scale and let the compositor scale things down — which means larger buffers, so extra VRAM usage and GPU downscaling overhead. I tested it with Steam and Minecraft so far, and it is fully stable.
3. I cleaned up the Russian comments.
4. Sorry about the word salad. English isn’t my native language, so I sometimes use LLMs to translate and organize my thoughts. I’ll keep the comments shorter and more direct.

ilia-eiq:
Would be nice to have a setting for base scale, e.g. if we had it set to 1, we would have "classic" xwayland behaviour.

gimgiwer:
> Would be nice to have a setting for base scale, e.g. if we had it set to 1, we would have "classic" xwayland behaviour.

Can you say a bit more about what you're trying to do? I'm not totally sure what you're after. if you just want the classic xwayland behavior, set XWAYLAND_SATELLITE_DISABLE_COMPOSITOR_SCALING=1 to turn off compositor scaling.

ilia-eiq:
@gimgiwer I mean on low end systems it would solve the resource usage you mentioned (vram and computing power to render at high resolution) while keeping windows scale consistent across monitors with different DPI.

gimgiwer:
> @gimgiwer I mean on low end systems it would solve the resource usage you mentioned (vram and computing power to render at high resolution) while keeping windows scale consistent across monitors with different DPI.

Added XWAYLAND_SATELLITE_BASE_SCALE (1.0–8.0). Set it to 1 and you're back to native res, VRAM stays low, and the positioning path is untouched so the geometry fixes still hold. Pretty much the classic behavior you were after.

PR #454, opened 2026-07-11
Stabilize re-anchoring when windows move across outputs 

sophietheopossum:
Stabilizes how windows are anchored to outputs, preventing X11 repositioning feedback loops that could occur during transient overlaps or layout changes.

-   Introduces a new component to track all outputs a surface is currently on.
-   Modifies the main anchoring component to be more stable, changing only when its output truly leaves or if the window isn't currently anchored to a valid output. This avoids excessive re-anchoring from transient overlaps.
-   Re-anchors windows to the most recently entered valid output when the current anchor output leaves.
-   Extracts the core window anchoring logic into a dedicated helper function, which handles X11 position updates and focused window primary output resets.
-   Ensures popups and other non-window surfaces continue to follow the most recently entered output, aligning with their short-lived nature and lack of X11 positions.
-   Updates the testing utility to accurately model surface leave events during output transitions, ensuring correct behavior when moving surfaces between outputs.

### DIFF
--- a/src/server/event.rs
+++ b/src/server/event.rs
@@ -277,11 +277,45 @@ impl SurfaceEvents {
                     return;
                 };
                 surface.leave(&output);
-                if data
+
+                if let Some(mut entered) = data.get::<&mut EnteredOutputs>() {
+                    entered.0.retain(|e| *e != output_entity);
+                }
+
+                let anchored_here = data
                     .get::<&OnOutput>()
-                    .is_some_and(|o| o.0 == output_entity)
-                {
-                    cmd.remove_one::<OnOutput>(target);
+                    .is_some_and(|o| o.0 == output_entity);
+                if anchored_here {
+                    // The anchor output is gone for real: re-anchor to the
+                    // most recently entered output the surface is still on.
+                    let next = data.get::<&EnteredOutputs>().and_then(|entered| {
+                        entered.0.iter().rev().copied().find(|e| {
+                            state
+                                .world
+                                .entity(*e)
+                                .is_ok_and(|en| en.has::<OutputDimensions>())
+                        })
+                    });
+                    match next {
+                        Some(next_output) => {
+                            let mut query = data.query::<(&x::Window, &mut WindowData)>();
+                            if let Some((window, win_data)) = query.get() {
+                                anchor_window_to_output(
+                                    &state.world,
+                                    *window,
+                                    win_data,
+                                    next_output,
+                                    &state.global_output_offset,
+                                    state.last_focused_toplevel,
+                                    connection,
+                                );
+                            }
+                            cmd.insert_one(target, OnOutput(next_output));
+                        }
+                        None => {
+                            cmd.remove_one::<OnOutput>(target);
+                        }
+                    }
                 }
             }
             Event::PreferredBufferScale { .. } => {}
--- a/src/server/event.rs
+++ b/src/server/event.rs
@@ -206,27 +206,43 @@ impl SurfaceEvents {
                 };
 
                 surface.enter(&output);
-                let on_output = OnOutput(output_entity);
 
                 debug!("{} entered {}", surface.id(), output.id());
 
+                // Track every output the surface is on. The anchor
+                // (OnOutput) only changes when its output actually leaves:
+                // re-anchoring on every enter turns transient overlaps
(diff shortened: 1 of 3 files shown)


En-En-Code:
Interesting. I already knew the `wl_surface::Enter` behavior is not correct, but I am not sure this fully resolves my problems with it. I'll tinker around with this and get back on this.

testwl/src/lib.rs
```diff
@@ -880,8 +882,37 @@ impl Server {
...
+            if old.id() != output.id() {
+                data.surface
+                    .leave(
+                        &old,
+                    );
+            }
+        }
+        if !already_entered {
+            data.surface.enter(output);
+        }
```
Supreeeme: This means `enter` will never be sent if the surface is exiting another output?

src/server/event.rs
```diff
@@ -240,26 +240,63 @@ impl SurfaceEvents {
...
                             OutputScaleFactor::Fractional(scale.0),
                         ) {
-                            state.updated_outputs.push(on_output.0);
+                            state.updated_outputs.push(output_entity);
                         }
                     }
+                } else {
+                    // Popups and other non-window surfaces keep the original
+                    // follow-the-latest-enter behavior; they are short-lived
+                    // and have no X11 position of their own to destabilize.
```
Supreeeme: Well this comment doesn't make much sense, popups are definitely window surfaces and do have X11 positions.

src/server/event.rs
```diff
@@ -240,26 +240,63 @@ impl SurfaceEvents {
...
             Event::Leave { output } => {
                 let output_entity = output.data().copied().unwrap();
-                let Ok(output) = state.world.get::<&WlOutput>(output_entity) else {
-                    return;
-                };
-                surface.leave(&output);
-                if data
+                if let Ok(output) = state.world.get::<&WlOutput>(output_entity) {
+                    surface.leave(&output);
+                }
```
Supreeeme: Unclear why this changed, if the entity doesn't have an output there's not much point in continuing.

Supreeeme (commented):
- The commit message for the first commit is not clear that it is specific to `testwl`.
- What exactly is this fixing, i.e. what applications? It's unclear to me how this is particularly different from the existing code.
- This needs a test.

sophietheopossum:
>     * What exactly is this fixing, i.e. what applications? It's unclear to me how this is particularly different from the existing code.

In current code, a window that overlaps two outputs' bounding boxes gets re-anchored to an output it only transiently entered. The X11 offset is rewritten to that output, so pointer coordinates no longer match where the window actually is: clicks land offset, and regions of the window stop responding entirely. Because repositioning can itself generate further enter events, it can settle into a feedback loop.

Reproducible with a compositor that sends wl_surface.enter on bounding-box overlap: in my case ShojiWM on a dual-output laptop, where a window straddling both screens would drift and misplace clicks.

How it differs: previously every enter set OnOutput and repositioned the window. Now the anchor only changes when it's invalid such as never set, or its output no longer has dimensions, so a window merely brushing another output keeps its anchor and its X11 position. EnteredOutputs exists solely so that when the anchor output does leave, there's a replacement to pick. Surfaces with no X11 window attached keep the old follow-the-latest-enter behaviour: either they haven't been associated with a window yet, or the window they had has since been destroyed. Either way there's no X11 position to keep stable.

Test will come soon once I've tidied up a few things

sophietheopossum:
>     * This needs a test.

test has been added, without this pr it fails

testwl/src/lib.rs
sophietheopossum: No, it only checks whether the surface was already on the target output, not whether it's leaving another. I've amended the commit to make that clearer.

src/server/event.rs
sophietheopossum: yeah my bad, popups are ordinary window surfaces here and do go through anchor_window_to_output. The branch is actually for surfaces with no X11 window attached: not yet associated via the serial handshake, or associated to a window that has since been destroyed. Fixed the comment to say that.

src/server/event.rs
sophietheopossum: oops meant to remove that change! it's gone now anyway.

sophietheopossum:
that should be everything now! please give it a look again and lmk if there's anything else you want clarified or changed

PR #455, opened 2026-07-11
add optional mode for mixed scaling factors

sophietheopossum:
this is entirely opt in. when enabled, there's still a global scaling factor for x windows, but on monitors with larger scaling factors windows will be scaled up. obviously this is a bit blurry, but in some use cases may be preferred over windows just completely ignoring scaling.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -51,6 +51,11 @@ pub trait RunData {
     fn max_req_len_bytes(&self) -> Option<usize> {
         None
     }
+    /// Present X11 clients with the compositor's logical coordinate space
+    /// instead of device pixels. See `ServerState::logical_geometry`.
+    fn logical_geometry(&self) -> bool {
+        false
+    }
 }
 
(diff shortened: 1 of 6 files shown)


En-En-Code:
I need to read the literature on how other compositors integrating `Xwayland` directly solve the mixed-DPI problem, because it definitely is not what we do, given the number of issues/PRs made to "fix" it. That said, one of `satellite`'s design philosophies is that it is opinionated, so ideally the right thing happens by default with no flags/environment variables needed. I would also expect this default solution to be minimally blurry, which will probably require I learn a bunch about rendering.

sophietheopossum:
yeah afaik you can just have it be large and blurry or small and crisp, not really any inbetween? but i agree it'd be nice if blurriness is avoidable

Supreeeme:
#452 is another alleged attempt at this, which, if it actually works properly, I would prefer.

sophietheopossum:
not sure how i missed #452 , I agree that's a better solution, will close in favour of that

PR #458, opened 2026-07-17
fix: resolve echoed popup configures from exact X11 geometry

Version33:
I was trying to get some VST3 plugins working in Bitwig via Yabridge, got most of them except ShaperBox3, popup menus on things like the hamburger menu didn't work. The popups would close instantly. Got frustrated and sent an AI on an adventure to fix it. It found that setting my display scale to 1.0 (from 1.25) fixes it, but I don't want to use 1.0 scaling.  
Looked further, found a bug in xwayland-satellite. I was skeptical, but applying the patch locally and testing the menu worked on my desired 1.25 scaling.

**The below description, code changes, and commit messages are all AI generated.** I tested the changes and it fixed my issue. A regression test was also added for this issue. 

```
Popup geometry currently round-trips X11 pixels -> logical (divide by
scale, truncate) -> X11 pixels (multiply by scale, truncate). At
fractional scales the double truncation is lossy: a popup requested at
parent-relative x=317 at scale 1.25 comes back as 316 (317/1.25 = 253.6
-> 253; 253*1.25 = 316.25 -> 316). The resulting off-by-one
ConfigureNotify move/resize is interpreted by strict clients as an
external change; JUCE popup menus (wine/yabridge audio plugins, Swing
menus behave the same) dismiss themselves instantly. Reproducible at
any non-integer scale, absent at 1.0.

Remember each popup's last positioner request (PopupData::requested):
the logical values sent to the compositor plus the exact X11 pixel
offset/size they were derived from. When a configure merely echoes the
granted request (logical values match), resolve the X11 geometry from
the stored pixels instead of rescaling. Configures the compositor
actually modified (constraint sliding etc.) keep the existing path
unchanged.
```

### DIFF
--- a/src/server/event.rs
+++ b/src/server/event.rs
@@ -300,17 +300,37 @@ impl SurfaceEvents {
             let (scale_factor, window, window_data, role) = query.get().unwrap();
 
             let window = *window;
-            let x = (pending.x.max(0) as f64 * scale_factor.0) as i32 + window_data.output_offset.x;
-            let y = (pending.y.max(0) as f64 * scale_factor.0) as i32 + window_data.output_offset.y;
-            let width = if pending.width > 0 {
-                (pending.width as f64 * scale_factor.0) as u16
+
+            // A configure exactly echoing a popup's positioner request resolves
+            // from its exact X11 pixel values (see PopupData::requested).
+            let exact = if let SurfaceRole::Popup(Some(p)) = &*role {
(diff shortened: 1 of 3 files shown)


Version33:
AI Used: Claude Fable 5
Harness: oh-my-pi
OS: NixOS
Config: https://github.com/Version33/dotfiles

PR #467, opened 2026-07-25
server: track global scale to initialize new surfaces correctly (test fixes)

En-En-Code:
Building upon #403. Resolves a concern I had with the integration tests hard-coding a 100x100 size for the configure event, even when the passed width and height values were different.

Also, a couple of questions about the rewrite.
1. How much scaling stuff is actually changing? One thing I realized about this change is that the scale could just be stored in the state with a flag for when it changes. Then, `OutputScaleFactor` and `SurfaceScaleFactor` can be removed from the ECS (or whatever data structure you're migrating to, where are you on this?) for being redundant data that also requires extra synchronization work.
2. Will `xcb` still be the X11 library post-rewrite? I did some work on a fix behind the scenes, but it does not make sense to polish up/write the tests for now if the library will be shortly changing.

### DIFF
--- a/src/server/dispatch.rs
+++ b/src/server/dispatch.rs
@@ -234,7 +234,7 @@ impl<S: X11Selection>
                         client,
                         server,
                         viewport,
-                        scale: SurfaceScaleFactor(1.0),
+                        scale: SurfaceScaleFactor(state.current_scale),
                     },
                 );
                 if let Some(f) = fractional {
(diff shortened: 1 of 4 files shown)


Supreeeme:
I'll probably merge #403 before this.
To answer the questions:
1. My goal is to avoid the ECS entirely and just store data as structs, basically using an async "actors" approach, basically what's described [in this article](https://ryhl.io/blog/actors-with-tokio/) (but I don't wanna use tokio)
2. I think I'm keeping xcb, for now.

En-En-Code:
That's fine. I branched off #403 to write this anyways. Normally I hate correcting other contributors' code but when they do not respond (or aren't using their brain at all), there's no contributor knowledge building to lose and it's easier to make the change yourself.

Supreeeme:
Merged #403, if you fix the merge conflicts I'll merge this.

Issue #256, opened 2025-10-23
Some WinApps windows quickly shrink


YaLTeR:
### Operating System

Fedora 43 Silverblue

### Wayland Compositor

niri, labwc

### Describe the issue.

[WinApps](https://github.com/winapps-org/winapps) lets you use Windows apps through some kind of xfreerdp window passthrough. Some windows works fine, but others, usually dialogs, rapidly shrink.

https://github.com/user-attachments/assets/cd589716-70de-4253-8c0c-4c53da73fa6a

This reproduces on xwayland-satellite on both niri and labwc. Labwc's native xwayland doesn't shrink those windows. But they appear a bit clipped on the bottom-right, like this Properties dialog here:

<img width="1248" height="945" alt="Image" src="https://github.com/user-attachments/assets/6be889af-fa1e-47fa-b3ec-7e4ecfaeb143" />

xwl-s log during the shrinking:

```
 2025-10-23T09:24:27.365Z DEBUG xwayland_satellite::xstate           > new window: CreateNotifyEvent { response_type: 16, pad: 1, sequence: 811, parent: Window { res_id: 838 }, window: Window { res_id: 4194428 }, x: 265, y: 394, width: 363, height: 499, border_width: 0, override_redirect: false, pad: 1 }
 2025-10-23T09:24:27.366Z WARN  xwayland_satellite::xstate           > unrecognized message: Atom { res_id: 281 }
 2025-10-23T09:24:27.366Z DEBUG xwayland_satellite::xstate           > requested to map Window { res_id: 4194428 }
 2025-10-23T09:24:27.367Z DEBUG xwayland_satellite::xstate           > requested to map Window { res_id: 4194428 }
 2025-10-23T09:24:27.367Z DEBUG xwayland_satellite::xstate           > Window { res_id: 4194428 } request: X | Y
 2025-10-23T09:24:27.368Z DEBUG xwayland_satellite::server           > setting Window { res_id: 4194428 } decorations Client
 2025-10-23T09:24:27.368Z DEBUG xwayland_satellite::xstate           > Window { res_id: 4194428 } window_types: ["_NET_WM_WINDOW_TYPE_DIALOG"]
 2025-10-23T09:24:27.368Z DEBUG xwayland_satellite::xstate           > Window { res_id: 4194428 } override_redirect: false
 2025-10-23T09:24:27.368Z DEBUG xwayland_satellite::server           > mapping Window { res_id: 4194428 }
[...]

PR #469, opened 2026-07-31
fix: pair of clientside decoration bugs

En-En-Code:
First is an overflow that occurs when adding the titlebar height to the min/max size hint.

Second is that bug where Weston shrinks its windows by the titlebar height whenever they lose focus. I found the source of this bug completely on accident lol.

### DIFF
--- a/src/server/event.rs
+++ b/src/server/event.rs
@@ -291,13 +291,8 @@ impl SurfaceEvents {
         drop(xdg);
 
         if let Some(pending) = pending {
-            let mut query = data.query::<(
-                &SurfaceScaleFactor,
-                &x::Window,
-                &mut WindowData,
-                &mut SurfaceRole,
-            )>();
-            let (scale_factor, window, window_data, role) = query.get().unwrap();
+            let mut query = data.query::<(&SurfaceScaleFactor, &x::Window, &mut WindowData)>();
+            let (scale_factor, window, window_data) = query.get().unwrap();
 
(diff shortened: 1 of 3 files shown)


src/server/event.rs
```diff
@@ -505,32 +499,33 @@ pub(super) fn update_surface_viewport(
...
     if let Some(hints) = size_hints {
-        let Some(data) = toplevel_data else {
-            return;
-        };
-
-        let decorations_height = if data.decoration.satellite.is_some() {
-            DecorationsDataSatellite::TITLEBAR_HEIGHT
-        } else {
-            0
+        if let Some(data) = toplevel_data {
```
Supreeeme, line 502:
```suggestion
if let Some(hints) = size_hints && let Some(data) = toplevel_data {
```

Supreeeme (commented):
I wonder if this would fix #256?

src/server/event.rs
En-En-Code: Even though 2024 edition (1.85.0+) landed `if let` rescoping, the feature does not let stabilized until 1.88.0, which for us is an MSRV bump. Let me know if you still want the change.

En-En-Code: Considering there are other places this refactor would apply, along with the condiitional `UnixStream`/`PipeWriter` construct in `testwl` that could be removed, I am inclined to make bumping MSRV its own PR.

Issue #438, opened 2026-06-07
Some steam games are starting inside steam client


9Prestidigitator:
### Operating System

NixOS

### Wayland Compositor

Niri

### Describe the issue.

Civilization V is starting inside the Steam client window on Niri.

The only workaround is by closing all x11 clients after launching the game. When doing that I get the following niri window:
```bash
Window ID 254:
  Title: "Sid Meier's Civilization V (DX11)"
  App ID: "steam_app_8930"
  Is floating: no
  PID: 3259
  Workspace ID: 3
  Layout:
    Tile size: 1920 x 1080
    Scrolling position: column 2, tile 1
    Window size: 1920 x 1080
    Window offset in tile: 0 x 0
```

### Please confirm you have reproduced this on the latest commit.

- [x] I have reproduced this on the latest commit

larsch:
I saw this exact issue also when I installed a new QWHD monitor (ARC Raiders). I removed window matching rules to put steam games on workspace 1 from my niri config, and the issue went away completely.

selfopenhell-byte:
This exact issue happens to me but with Forza Horizon 6, the game gets stuck in the steam client and doesn't make a new window that niri recognizes, the glued window is still identified as AppID: ''steam'' via the command ''niri msg windows''.

Most game launchers also launch within the steam client for me, even when the actual game launches in its own window the vast majority of the time.

Edit: 2026-06-18
OS: Fedora 44
Compositor: Niri

ArikRahman:
This happens to me on Warframe with Nixos Niri. I have to close the client and put the game on another workspace to start it correctly.

larsch:
I've been investigating the "Steam games appearing inside the Steam client" issue from #438, along with related fullscreen/resize problems at fractional scales.

**Root cause:** Multiple scale-related bugs, all stemming from the same underlying issue — xwayland-satellite wasn't correctly handling the difference between physical and logical pixel sizes on fractional-scale outputs.

**The three fixes in this branch:**

1. **Size hint rounding (`ceil()` instead of `floor`)** — When a game pins its window size via `WM_NORMAL_HINTS` with `min_size == max_size` and then goes fullscreen, the X11 pixel values cycle through a floor-truncation feedback loop, losing ~1px per round-trip. The window shrinks below fullscreen and the compositor exits the fullscreen state. Fixed by using `ceil()` when converting X11 pixel constraints to logical coordinates, matching how the viewport is already handled.

[...]

Issue #428, opened 2026-05-02
Cursor position getting misreported in RuneScape3 Client after selecting other X11 application


WerWolv:
### Operating System

Fedora 44

### Wayland Compositor

Niri

### Describe the issue.

Any cursor inputs to the RuneScape 3 client get mis-reported after focusing another X11 application until the RuneScape 3 client is resized / moved. So far this was the only X11 application where I could observe this behaviour but it's very consistent there.

RuneScape changes the cursor icon depending on what you're hovering over in-game. When in this broken state, the cursor is correctly rendered where the mouse is at but it has the icon of the thing the game thinks it hovers over. Clicking sends the input to that wrong position.

### Steps to reproduce

- Start RuneScape 3
    - It can be downloaded for free from Steam and should just run like this through Proton: https://store.steampowered.com/app/1343400/RuneScape/
    - It's most noticeable and annoying when in game of course but simply being on the login-screen is enough to see it.
- Open a second application that also uses X11 and interact with it in any way (clicking on it or just focusing it)
    - The Steam Client itself or any other application that uses X11, it seems to happen with any
- Go back to the RuneScape 3 client and try to interact with it
- Often the first click gets registered correctly but after that any clicks to e.g the right side of the window get reported to the left side of the Game. Same with hovering over things. The cursor changes its icon depending on where the game thinks the cursor is at, not where it actually is at
- It can be resolved in Niri by grabbing the Window and moving it around (e.g getting it to resize and dock elsewhere on the screen and then full-screening it again). This will fix it until the next time another X11 application is being focused

### Please confirm you have reproduced this on the latest commit.

- [x] I have reproduced this on the latest commit

WerWolv:
So, full disclosure I don't know enough about X11 to know the impact of this and I used AI to make this change.

Removing this entire if-else block here makes RuneScape work correctly: 
https://github.com/Supreeeme/xwayland-satellite/blob/main/src/xstate/mod.rs#L1402-L1429

Issue #385, opened 2026-02-23
Issue: Input (Mouse/Keyboard) Loses Focus When Switching Monitors with Multiple EVE Instances


Mathew-D:
### Operating System

Manjaro

### Wayland Compositor

Niri

### Describe the issue.

When running four separate EVE Online windows via xwayland-satellite (one on each monitor), the mouse and keyboard stop responding on the newly focused monitor unless I toggle fullscreen/unfullscreen on the current window. Toggling fullscreen restores input, but then it breaks again when I move to a different monitor.

This makes it impossible to smoothly switch focus between instances.

### Steps to reproduce

1. Launch 4 instances of EVE Online, one on each monitor (via xwayland-satellite).
3. Ensure each instance is visible and running.
5. Move mouse focus to a different monitor/window.
7. Try clicking / using keyboard in the newly focused window.
9. Observe that input does not get delivered to that window.
11. Toggle fullscreen on the current monitor’s window.
13. Input returns and works fine until you switch to another monitor — then the cycle repeats.

### Please confirm you have reproduced this on the latest commit.

- [x] I have reproduced this on the latest commit

rykugur:
I'm also experiencing this, in both EVE Online and several other games. I made an Issue in the Niri repo thinking it had to do with Niri (since I didn't experience it in Hyprland), but they helped determine that it was likely due to xwayland-satellite (and then I found this issue from four days ago).

https://github.com/niri-wm/niri/issues/3296

EDIT: moving some details here from the above issue.

This only seems to happen when crossing a monitor boundary.

Simple video showing what's going on: https://youtube.com/shorts/oOONElwbK3Y. It's not clear from the video, but when you see the camera quickly spin after I've clicked back on EVE to give it focus, it will not accept any keyboard/mouse input until I un-focus/refocus one or more times. Note: the camera spin should not be happening either.

Seralyne:
I'm experiencing similar issues. Originally from a different issue that I was asked if it still happening (which it wasn't), but I came to the conclusion from Rykugur's own issue that since xorg-xeyes is responsive, the response I'm gonna get is "It's not Niri, it's xwayland-sattelite". It's not quite the same issue as described above, but I believe it might be similar.

https://github.com/niri-wm/niri/issues/1959#issuecomment-4003054173. Here's the important bits from that:

[...]

PR #475, opened 2026-08-09
Fix persistent cursor offset in Proton games on multi-monitor setups

### LINKED ISSUE #321: Final Fantasy XIV (via XIVLauncher) has mouse issues with multimonitor setups
### Operating System

Gentoo

### Wayland Compositor

Niri

### Describe the issue.

When I focus a window on my 2nd monitor and then try to focus Final Fantasy XIV again my cursor shows the system cursor (instead of the FFXIV one) and is unable to interact with the window. The only fix I've found is using the "focus-monitor-next" bind in Niri to switch focus between monitors until it works again. It seems like it's actually a 50/50 chance each time the FFXIV window regains focus that my mouse will properly respond in the game.

If I'm only interacting with windows on the same monitor as FFXIV, it seems to work fine. I only notice the issue when I use "focus-monitor-next" to switch to a window on my other monitor and then try to click back into FFXIV.

Niri's Issue placeholder text told me to report issues with xwayland applications here, so I'm here, although this sounds like a Niri bug.

Application Details:
XIVLauncher 1.2.1 flatpak from Flathub (XIVLauncher only recognizes the flatpak as an official build last I checked)
Wine 10.8 bundled in XIVLauncher (https://github.com/goatcorp/wine-xiv-git/releases/tag/10.8.r0.g47f77594)

System Details:
Gentoo
Niri 25.11-r1
Tried xwayland-satellite 0.8 and commit f0ad674b7009a6afd80cea59d4fbf975dd68ee95 which seem to be basically identical code-wise?
Mesa 25.2.7 (Although since XIVLauncher is a flatpak it might be using whatever the flathub Mesa currently is)

AMD RX 9070 XT

### Steps to reproduce

Open a window on a secondary monitor
Launch FFXIV via XIVLauncher onto the main monitor
Use the Niri bind "focus-monitor-next" to focus a window on another monitor
Try to click back into FFXIV and notice that you cannot click on anything in the game

Hitting the "focus-monitor-next" bind fixes it sometimes. Seems to work 50% of the time.

### LINKED ISSUE #339: In game cursor breaks when switching to steam on a 2nd monitor and back to the game on main monitor
### Operating System

Arch

### Wayland Compositor

Niri

### Describe the issue.

https://github.com/user-attachments/assets/7cb5c246-ae83-4649-a3a1-b04a7223b29b

(Lower wall caused by monitor position settings, setting this to 0x0 fixes this but still cant move cursor back to the left) (game shown is warframe however this issue can be replicated on any other game i tested on with varying degrees of severity such as arc raiders and baldurs gate 3, this is the most extreme example)

My main monitor is 1440p and 2nd monitor is 1080p and is to the left of my main monitor, when switching to steam on my 2nd monitor and back to my game my in game cursor gets locked between the 1920 to 2560 pixel range, this does not happen with any other software I've tested on my 2nd monitor

The actual mouse position on the main monitor outside the game seems to be tracked as normal but in game the mouse position is desynced. i believe this is caused by  xwayland satellite not knowing how to handle 2 xwayland apps on different monitors/resolutions as when steam is also on my main monitor and i scroll back and forth this issue does not present itself

Matching the monitors resolution and scaling doesn't fix this

Neither does setting both monitors to x=0 y=0 

Toggling back and forth from fullscreen does fix the desync until focusing steam again

Gamescope and PROTON_ENABLE_WAYLAND=1 does prevent this issue but those also prevent steam overlay from running in my experience

### Steps to reproduce

1. Have steam open on a 2nd monitor and a game on the main monitor
2. Tab out into steam and back into the game
3. Cursor generally breaks here or after scrolling away and back to steam on the 2nd monitor 

larsch:
Two related fixes for Proton/Wine games losing cursor alignment when switching between monitors.

### 1. Set `_NET_DESKTOP_GEOMETRY` and `_NET_WORKAREA` on the root window

Wine queries these to determine monitor layout. When missing, it falls back to Xinerama with the warning:

```
warn:x11drv:get_work_area _NET_WORKAREA is not supported, Work areas may be incorrect.
```

The properties are now set at startup and updated when outputs or the global output offset change. Each output's full geometry is reported as a work area; `_NET_DESKTOP_GEOMETRY` is the bounding box of all outputs.

### 2. Keep the primary output stable

The root cause of the cursor offset: `SetOutputPrimary` was called on every focus change between monitors. This rotates the Xinerama coordinate origin — when a different monitor becomes primary, the entire X11 → Windows desktop coordinate mapping shifts by the monitor width. Proton games that cache their window position relative to the old origin compute cursor coordinates off by exactly one screen width.

With this fix, the primary output is set *once* on the first focus event, then never changed. The Xinerama coordinate system remains stable regardless of which monitor has focus.

**Test:** Updated `primary_output` integration test to verify the primary remains on the first-focused output after focus changes and output hotplug.

Fixes #339
Likely fixes #321, #66, #428, #385
Ref: #438

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 /target
 /result
+.worktrees/
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -23,6 +23,7 @@ pub trait XConnection: Sized + 'static {
     fn close_window(&mut self, window: x::Window);
     fn unmap_window(&mut self, window: x::Window);
     fn raise_to_top(&mut self, window: x::Window);
+    fn update_desktop_properties(&self, _outputs: &[(i32, i32, i32, i32)]) {}
 }
 
 pub trait X11Selection {
--- a/src/server/mod.rs
+++ b/src/server/mod.rs
@@ -656,6 +656,8 @@ impl<C: XConnection> ServerState<C> {
                 .unwrap();
         }
 
+        let mut outputs_changed = false;
+
         if self.global_output_offset.x.owner.is_none()
             || self.global_output_offset.y.owner.is_none()
         {
@@ -663,6 +665,7 @@ impl<C: XConnection> ServerState<C> {
             self.global_offset_updated = true;
         }
         if self.global_offset_updated {
+            outputs_changed = true;
             debug!(
                 target: "output_offset",
                 "updated global output offset: {}x{}",
@@ -681,6 +684,7 @@ impl<C: XConnection> ServerState<C> {
         }
 
         if !self.updated_outputs.is_empty() {
+            outputs_changed = true;
             for output in std::mem::take(&mut self.updated_outputs).iter() {
                 let Ok(output_scale) = self.world.get::<&OutputScaleFactor>(*output) else {
                     continue;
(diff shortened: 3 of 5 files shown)


Kyedae:
This helped me so much, thank you

MaveDX:
Can confirm this works, havent had any issues with it in the last few days ive been using it

En-En-Code:
Neat to know this is also a bug caused by primary output swapping. I tested switching primary outputs in `i3` and it also experiences the cursor cut-offs. We know this behavior is wrong--the problem is determining how primary should be managed exactly (see the discussion in #442).

PR #476, opened 2026-08-11
feat: syslog support

En-En-Code:
Here is the `syslog` version of #414. Haven't updated the documentation yet (I'm tired).

Yes, the panic hook is important. Unwinding is considered "normal" process termination and is not caught by `coredumpctl` or similar utilities (unlike, for example, segfaulting).

I'll probably ask users with `systemd-journald` something like:
`journalctl --grep=xwayland-sat --output=short-iso-precise -b > "$(date -Iseconds).log"`.
Users without `systemd` will have to figure out where in `/var/log` their syslog daemon outputs to, but that shouldn't be too hard to explain.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yaml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yaml
@@ -20,8 +20,16 @@ body:
     id: problem
     attributes:
       label: Describe the issue.
-      value: |
-        **Please attach a log if applicable.**
+      description: |
+        Provide a debug log file.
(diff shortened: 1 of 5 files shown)


Supreeeme:
Do you have any preferences between logging to syslog vs a file? I am leaning towards syslog.

En-En-Code:
I think `syslog` is the better approach, since debug mode produces enough logs to where rotating them is actually nice. It has a few caveats, like not being able to get full panic backtraces, but as long as issue creators can Read the Fabulous Directions :tm:, it is still a great improvement over our current status quo.

Issue #130, opened 2025-03-24
Windows running remote servers desktop hang after exiting program


lukeflo:
### Operating System

Ubuntu 24.04. based Tuxedo OS

### Wayland Compositor

Niri

### Describe the issue.

I'm using Niri as WM and xwayland-satellite as simple and good working. Part of my work is connecting to our companies remote servers desktop through `x2goclient` with the [Wayland-trick](https://wiki.archlinux.org/title/X2Go#X2Go_client:_Cannot_connect_to_remote_X2Go_server_from_local_Wayland_session:_%22The_connection_with_the_remote_server_was_shut_down%22) and doing stuff on the server machine. Using my preinstalled KDE Desktop, which uses Xwayland directly, everything just works. But with Niri and Satellite I run into some problems.

I can connect to the server through x2go and access the desktop. But after logging out from the remote desktop, the window in which the desktop ran and the window containing the x2goclient just don't close. Plus they are not responding to any action. I even can't close them with my Niri shortcut for exiting a window. When grepping for the x2go processes via `pgrep`, I can see that the processes themselves are gone. Only the windows remain in a frozen state.

The only way to close them is to kill the whole xwayland-satellite. Of course, this is no long-term solution.

The generated log output from starting x2goclient until exiting the remote desktop and the hanging starts ist:

```bash
 2025-03-24T14:50:37.916Z WARN  xwayland_satellite::server::event > unhandled xdgtoplevel event: ConfigureBounds { width: 1904, height: 1034 }
 2025-03-24T14:50:37.916Z WARN  xwayland_satellite::server::event > unhandled xdgtoplevel event: WmCapabilities { capabilities: [3, 0, 0, 0] }
 2025-03-24T14:50:56.051Z WARN  xwayland_satellite::xstate        > unrecognized message: Atom { res_id: 252 }
 2025-03-24T14:50:56.052Z WARN  xwayland_satellite::server::event > unhandled xdgtoplevel event: ConfigureBounds { width: 1920, height: 1050 }
 2025-03-24T14:50:56.052Z WARN  xwayland_satellite::server::event > unhandled xdgtoplevel event: WmCapabilities { capabilities: [3, 0, 0, 0] }
 2025-03-24T14:50:56.055Z WARN  xwayland_satellite::xstate        > unrecognized message: Atom { res_id: 252 }
 2025-03-24T14:51:13.565Z INFO  xwayland_process                  > syntax error: line 1 of stdin
[...]

PR #484, opened 2026-08-22
Drop parent copy of the Xwayland WM socket after spawn

am5083:
Refs #130.

## Problem

xwayland-satellite creates a socket pair for communication with
Xwayland. It passes `xsock_xwl` to Xwayland through `-wm`, but the
satellite process retains its own copy after spawning Xwayland.

A Unix socket reports EOF only after all references to 
the peer end are closed. If Xwayland exits, the retained 
copy prevents `xsock_wl` from reporting EOF.

If `satellite` is in a blocking xcb request when Xwayland exits, that request can hang indefinitely. `Satellite` then does not return to the main poll loop to read the Xwayland exit status. Its Wayland toplevels remain visible as frozen windows.

## Change

Drop `xsock_xwl` after spawning Xwayland, next to the
`drop(fds)` call.

## Verification

- `cargo fmt --check`
- `cargo build --all-targets --profile ci --locked`
- `cargo test --profile ci --locked -- --test-threads 1 --nocapture`
- `cargo clippy --all-targets --profile ci --locked --workspace`

All pass

## Notes

This change makes a dead Xwayland connection observable by libxcb.
Some dead-connection paths currently panic instead of shutting down cleanly, 
including `poll_for_event().unwrap()` in `handle_events`.

This issue is not addressed in this PR because it is a larger behavioral change.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -110,6 +110,8 @@ pub fn main(mut data: impl RunData) -> Option<()> {
 
     // Now that Xwayland spawned and got the listenfds, we can close them here.
     drop(fds);
+    // Close our copy of the WM socket so Xwayland exiting produces EOF on xsock_wl.
+    drop(xsock_xwl);
 
     let xwl_pid = xwayland.id();
 


En-En-Code:
Neat. I was expecting the issue to be resolved with the async rewrite having an async mainloop that can exit on `Xwayland`'s termination, but this should work too.

I'm not concerned about the `unwrap` on `poll_for_event` since there is nothing better for `satellite` to do than terminate if it cannot communicate with `Xwayland` anymore, and `unwrap` is the most concise way to do that.

En-En-Code (approved):

am5083:
Nice. Failing `cargo fmt --check` looks like it comes from `test/integration.rs` on `main`

PR #494, opened 2026-09-06
fix steam dropdowns closing instantly and unity add component not focusing

### LINKED ISSUE #468: Drop down selection menu fail to render after v0.8.2 (commit 3273a0f)
### Operating System

CachyOS

### Wayland Compositor

Niri

### Describe the issue.

I use Niri and I run Assetto Corsa Content Manager via Steam under GE-Proton 9-20 as usual. But after upgrading to v0.8.2, all drop down selection menu will not render, click on them just like nothing happen, e.g. selecting car colors, profiles, presets, etc ... all stopped rendering under v0.8.2. But when I downgrade to v0.8.1, they will render normally.

Also I tested out v0.8.2 under KDE Plasma, all drop down menus are rendering normally. Since KDE does not use xwayland-satellite, this bug doesn't exists at all.

My current solution is to pin to v0.8.1 to make this game working again under Niri.

### Steps to reproduce

I check out each commit under v0.8.2, compile each commit one by one to test each of them, and here are the results:
  * `a879e5e` - Working normally
  * `3273a0f` - Broken drop down menu
  * `5d1efbc` - Broken drop down menu

So for sure the problem is from commit `3273a0f`. I am not sure about the technical details in it, but if this commit is all about fixing unity editor component pop up, it may have also cause some WPF/.NET applications drop down menu failed to render.

3akev:
# Summary

This PR fixes #468 and #278 by implementing more comprehensive popup management that takes into account `override_redirect` and the `WM_TAKE_FOCUS` atom.

In short, the new algorithm is as follows:
- If `override_redirect` is set, never focus.
- otherwise, if `WM_TAKE_FOCUS` is present, send a ClientMessage containing `WM_TAKE_FOCUS`, letting the client choose which window to focus.
- otherwise, set focus to the window in question. 

This behaviour fixes both bugs and seems to be more or less consistent with what is done by other compositors, for example [Hyprland](https://github.com/hyprwm/Hyprland/blob/34eb03bd8da01024596c367fba66485a8c9b8ca7/src/xwayland/XWM.cpp#L546-L581)

## Testing

Confirmed to be working with the latest versions of steam and unity on Niri, running on Arch.
Unit tests have been added to this effect, using the window properties of both popups:

### Steam:
```
xwininfo: Window id: 0x1e00097 (has no name)
[...]
  Override Redirect State: yes
```
```
WM_PROTOCOLS(ATOM): protocols  WM_DELETE_WINDOW, _NET_WM_PING
_NET_WM_WINDOW_TYPE(ATOM) = _NET_WM_WINDOW_TYPE_POPUP_MENU
[...]
WM_HINTS(WM_HINTS):
		window id # of group leader: 0x16655abc
```
In short, `override_redirect=true, accepts_input=true, take_focus=false`. It is never focused.

### Unity:
```
xwininfo: Window id: 0x60a489 "UnityEditor.AddComponent.AddComponentWindow"
[...]
  Override Redirect State: no
```
```
WM_HINTS(WM_HINTS):
		Initial state is Normal State.
		bitmap id # to use for icon: 0x60a48d
		bitmap id # of mask for icon: 0x60a493
		window id # of group leader: 0x600001
[...]
_NET_WM_WINDOW_TYPE(ATOM) = _NET_WM_WINDOW_TYPE_POPUP_MENU
[...]
WM_PROTOCOLS(ATOM): protocols  WM_DELETE_WINDOW, WM_TAKE_FOCUS, _NET_WM_SYNC_REQU
```
In short, `override_redirect=false, accepts_input=true, take_focus=true`. It is offered focus via `WM_TAKE_FOCUS`.

### DIFF
--- a/src/server/event.rs
+++ b/src/server/event.rs
@@ -438,7 +438,7 @@ impl SurfaceEvents {
                             window,
                             output_name: None,
                             is_popup: true,
-                            is_take_focus: window_data.attrs.has_take_focus,
+                            has_take_focus: window_data.attrs.has_take_focus,
                         });
                     }
                 }
@@ -866,12 +866,12 @@ impl Event for client::wl_keyboard::Event {
                 ));
                 let output_name = get_output_name(output, &state.world);
                 let window_data = data.get::<&WindowData>();
-                let is_take_focus = window_data.as_ref().is_some_and(|d| d.attrs.has_take_focus);
+                let has_take_focus = window_data.as_ref().is_some_and(|d| d.attrs.has_take_focus);
                 state.to_focus = Some(FocusData {
                     window: *window,
                     output_name,
                     is_popup: false,
-                    is_take_focus,
+                    has_take_focus,
                 });
                 keyboard.enter(serial, surface, keys);
             }
--- a/src/server/event.rs
+++ b/src/server/event.rs
@@ -431,13 +431,14 @@ impl SurfaceEvents {
                 });
 
                 if first_configure {
-                    let window_data = data.get::<&WindowData>().unwrap();
-                    if window_data.attrs.require_wm_focus() {
-                        let window = *data.get::<&x::Window>().unwrap();
+                    let attrs = &data.get::<&WindowData>().unwrap().attrs;
+                    let window = *data.get::<&x::Window>().unwrap();
+                    if !attrs.override_redirect && (attrs.has_take_focus || attrs.accepts_input) {
                         state.inner.to_focus = Some(FocusData {
                             window,
                             output_name: None,
                             is_popup: true,
+                            is_take_focus: attrs.has_take_focus,
                         });
                     }
                 }
@@ -864,10 +865,17 @@ impl Event for client::wl_keyboard::Event {
                     serial,
                 ));
                 let output_name = get_output_name(output, &state.world);
+                let window_data = data.get::<&WindowData>();
+                let is_take_focus = if let Some(data) = window_data {
+                    data.attrs.has_take_focus
+                } else {
+                    false
+                };
                 state.to_focus = Some(FocusData {
                     window: *window,
                     output_name,
                     is_popup: false,
+                    is_take_focus,
                 });
                 keyboard.enter(serial, surface, keys);
             }
--- a/src/server/event.rs
+++ b/src/server/event.rs
@@ -431,14 +431,14 @@ impl SurfaceEvents {
                 });
 
                 if first_configure {
-                    let attrs = &data.get::<&WindowData>().unwrap().attrs;
-                    let window = *data.get::<&x::Window>().unwrap();
-                    if !attrs.override_redirect && (attrs.has_take_focus || attrs.accepts_input) {
+                    let window_data = data.get::<&WindowData>().unwrap();
+                    if window_data.attrs.require_wm_focus() {
+                        let window = *data.get::<&x::Window>().unwrap();
                         state.inner.to_focus = Some(FocusData {
                             window,
                             output_name: None,
                             is_popup: true,
-                            is_take_focus: attrs.has_take_focus,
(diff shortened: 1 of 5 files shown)


HcgRandon:
Running this locally; confirmed fully fixes the steam drop down issue.

src/server/event.rs
```diff
@@ -431,13 +431,14 @@ impl SurfaceEvents {
                 });
 
                 if first_configure {
-                    let window_data = data.get::<&WindowData>().unwrap();
-                    if window_data.attrs.require_wm_focus() {
-                        let window = *data.get::<&x::Window>().unwrap();
+                    let attrs = &data.get::<&WindowData>().unwrap().attrs;
+                    let window = *data.get::<&x::Window>().unwrap();
+                    if !attrs.override_redirect && (attrs.has_take_focus || attrs.accepts_input) {
```
Supreeeme: This condition should just go inside of `require_wm_focus` and this area remains unchanged.

src/server/event.rs
```diff
@@ -864,10 +865,17 @@ impl Event for client::wl_keyboard::Event {
                     serial,
                 ));
                 let output_name = get_output_name(output, &state.world);
+                let window_data = data.get::<&WindowData>();
+                let is_take_focus = if let Some(data) = window_data {
+                    data.attrs.has_take_focus
+                } else {
+                    false
+                };
```
Supreeeme:
```suggestion
                let is_take_focus = window_data.as_ref().is_some_and(|d| d.attrs.has_take_focus);
```

src/server/tests.rs
```diff
@@ -1647,104 +1707,44 @@ fn override_redirect_choose_hover_window() {
 }
 
 #[test]
-fn popup_no_focus_without_input_hint() {
```
Supreeeme, line 1650: I'm not sure about deleting this test.

src/server/tests.rs
```diff
@@ -1647,104 +1707,44 @@ fn override_redirect_choose_hover_window() {
...
+                f.connection().send_take_focus_window,
+                None,
+                "accepts_input={accepts_input:?} take_focus={take_focus}"
+            );
+        }
+    }
 }
 
 #[test]
-fn popup_focus_on_map_with_input_hint() {
```
Supreeeme, line 1666: Not sure about this one either.

Supreeeme:
Also I'm not sure if you added that flyover thing or if someone else did but please never do that again.

src/server/event.rs
3akev: Done

src/server/tests.rs
3akev: Re-added.

src/server/tests.rs
3akev: This test captures a case where override_redirect=true, which is now covered by the test `popup_override_redirect_never_focused_nor_offered`. The behaviour it tests changes with this PR anyways.

3akev:
Thanks for the review! I've made some changes, let me know if there's anything else.

FWIW I don't know what flyover is either.

src/server/mod.rs
```diff
@@ -398,6 +399,7 @@ struct FocusData {
     window: x::Window,
     output_name: Option<String>,
     is_popup: bool,
+    is_take_focus: bool,
```
Supreeeme:
```suggestion
    has_take_focus: bool,
```

Supreeeme (approved):

src/server/mod.rs
3akev: Done

ItsLemmy:
Just my two cents, but IMO this warrants a new release.  So many peoples are disoriented and frustrated by this one.

RomainLanz:
Hi there! 👋🏻 

When can we expect the next release?

Issue #472, opened 2026-08-03
Unusual cursor offset under simple SDL app.


sealyium:
### Operating System

Debian 13

### Wayland Compositor

Niri

### Describe the issue.

So, here i am again, i was experimenting with SDL2, and as a test i tried to draw something to the cursor.
But it had a weird offset that got bigger and bigger the more i went down the y axis.

It seems like this occurs only after using xwayland-sattelite, i tried the exact same program by forcing wayland : 
SDL_setenv("SDL_VIDEODRIVER", "wayland", 1);

I also ran the same thing on xfce with native x11.

And the offset doesn't appear for those two tests.

<img width="716" height="1131" alt="Image" src="https://github.com/user-attachments/assets/06db7f01-4220-4dd8-8e0b-307c93c08a79" />

### Steps to reproduce

Here is a trimmed down version of the sdl program : 

#include <SDL2/SDL.h>
#include <stdio.h>
#include <stdbool.h>

int main(void)
{
    if (SDL_Init(SDL_INIT_VIDEO) != 0) {
        fprintf(stderr, "%s\n", SDL_GetError());
        return 1;
    }

    SDL_Window *window = SDL_CreateWindow(
        "SDL Mouse Test",
        SDL_WINDOWPOS_CENTERED,
        SDL_WINDOWPOS_CENTERED,
        900,
        900,
        SDL_WINDOW_RESIZABLE);

    SDL_Renderer *renderer =
        SDL_CreateRenderer(window, -1, SDL_RENDERER_ACCELERATED);

    if (!renderer)
        renderer = SDL_CreateRenderer(window, -1, SDL_RENDERER_SOFTWARE);

    printf("Video driver : %s\n", SDL_GetCurrentVideoDriver());

    int ww, wh;
    int rw, rh;
    float sx, sy;

    SDL_GetWindowSize(window, &ww, &wh);
    SDL_GetRendererOutputSize(renderer, &rw, &rh);
    SDL_RenderGetScale(renderer, &sx, &sy);

    printf("Window   : %d x %d\n", ww, wh);
    printf("Renderer : %d x %d\n", rw, rh);
    printf("Scale    : %.2f %.2f\n", sx, sy);

    bool quit = false;
    int mx = 0, my = 0;

    while (!quit) {

        SDL_Event e;

        while (SDL_PollEvent(&e)) {

            switch (e.type) {

            case SDL_QUIT:
                quit = true;
                break;

[...]

Issue #347, opened 2026-01-11
satellite in Weston: Shrinking windows on focus switch


mviereck:
### Operating System

Debian trixie

### Wayland Compositor

Weston

### Describe the issue.

If I run xwayland-satellite in Weston, open satellite windows reduce their height every time I switch focus between different windows. It doesn't matter if the other windows are Wayland or X clients.
It is always about the height of the window title that gets lost.
(This does not happen in gnome-shell.)

(Not important for me, in Weston I set up a rootless Xwayland myself. Just want to note so you are aware.)

### Steps to reproduce

 - Run xwayland-satellite in Weston
 - Open some windows.
 - Switch focus between windows

### Please confirm you have reproduced this on the latest commit.

- [x] I have reproduced this on the latest commit

Supreeeme:
This is probably related to how the satellite rendered decorations work. Does Weston have server side decorations?

mviereck:
> Does Weston have server side decorations?

I am not sure. AFAIK no.

If I run rootful Xwayland with option `-decorate`, the Xwayland window has a decoration. I know that the libdecor package with cairo plugin is needed to allow Xwayland window decoration.
(Edit: works for rootless Xwayland as well.)

Supreeeme:
Yeah, then this is probably related to decorations. Can you run satellite with `RUST_LOG=debug` set and post the log?

mviereck:
I did:
start Weston
start satellite
open xterm window
switch a few times between Weston terminal and xterm (-> xterm window is shrinking)
```
 2026-01-11T18:22:16.553Z INFO  xwayland_satellite > Starting xwayland-satellite version v0.7-70-gd06dc15
 2026-01-11T18:22:16.559Z WARN  xwayland_satellite::server > Couldn't bind fractional scale manager: the requested global was not found in the registry. Fractional scaling will not work.
 2026-01-11T18:22:16.559Z WARN  xwayland_satellite::server > Could not bind xdg activation (NotPresent). Windows might not receive focus depending on compositor focus stealing policy.
 2026-01-11T18:22:16.559Z INFO  xwayland_satellite::server::selection > Primary selection unsupported.
 2026-01-11T18:22:16.591Z DEBUG xwayland_satellite::server            > Using new scale 1
 2026-01-11T18:22:16.591Z DEBUG xwayland_satellite::server::event     > moving wl_output@11[0] to 0x0
[...]

PR #496, opened 2026-09-06
fix: keep viewport in sync with X window size for satellite decorations

mdsitton:
Commit d1e2209 moved the titlebar height reduction from the configure handler into update_surface_viewport. The X window then kept the full configured height while the viewport destination was made TITLEBAR_HEIGHT shorter, so every toplevel with satellite drawn decorations was squished vertically. Pointer coordinates are only scaled by the output scale, so clicks drifted from the visible content, by up to 25px at the bottom of the window.

Subtract the titlebar height from the X window size again when resolving a configure, but only when the configure carries a non-zero height. A zero height keeps the stored size, which already excludes the titlebar, so windows no longer shrink on the zero-height configures that caused #347. The viewport is once more exactly the X window size.


src/server/event.rs
```diff
@@ -479,6 +478,16 @@ impl SurfaceEvents {
...
+/// Converts a size in X pixels to the surface's logical size at the given scale, rounding up so
+/// that the whole buffer is shown. Everything deriving a logical size from an X window size must
+/// use this so that viewport and popup geometry agree.
+pub(super) fn logical_size(width: u16, height: u16, scale: f64) -> (i32, i32) {
+    (
+        (width as f64 / scale).ceil() as i32,
+        (height as f64 / scale).ceil() as i32,
+    )
+}
+
```
En-En-Code: Do we really need this helper? It might be nice if there were a newtype pattern wrapping phyiscal and logical sizes, but there is not, and such a refactor should be a seperate PR.

src/server/tests.rs
```diff
@@ -3147,6 +3147,49 @@ fn client_side_decorations() {
...
+    // The titlebar is subtracted in logical pixels, so the X window content and the viewport
+    // agree at fractional scales as well (scale, preferred_scale value, expected X height).
+    for (scale, preferred, x_height) in [(1.5, 180, 112), (1.25, 150, 93)] {
+        let mut f = TestFixture::new_pre_connect(|testwl| {
+            testwl.enable_fractional_scale();
+        });
+        let compositor = f.compositor();
+        let (_, output) = f.new_output(0, 0);
+        let window = Window::new(1);
+        let (_, id) = f.create_toplevel(&compositor, window);
```
En-En-Code: Push outsize the loop.

src/server/tests.rs
```diff
@@ -3147,6 +3147,49 @@ fn client_side_decorations() {
     assert!(!subsurface.is_alive());
 }
 
+#[test]
+fn client_side_decorations_fractional_scale() {
+    // The titlebar is subtracted in logical pixels, so the X window content and the viewport
+    // agree at fractional scales as well (scale, preferred_scale value, expected X height).
+    for (scale, preferred, x_height) in [(1.5, 180, 112), (1.25, 150, 93)] {
```
En-En-Code: `x_height` feels too much like a magic number here. I would like to see it calculated inside the loop.

src/server/tests.rs
```diff
@@ -3147,6 +3147,49 @@ fn client_side_decorations() {
...
+            .preferred_scale(preferred);
+        f.testwl.move_surface_to_output(id, &output);
+        f.run();
+        f.run();
+
+        f.testwl
+            .force_decoration_mode(id, zxdg_toplevel_decoration_v1::Mode::ClientSide);
+        f.testwl.configure_toplevel(id, 100, 100, vec![]);
+        f.run();
+        f.run();
```
En-En-Code: Can probably also be moved outside the loop.

src/server/tests.rs
```diff
@@ -3233,6 +3237,15 @@ fn resize_decorations_on_reconfigure() {
...
 
+    // A resize from the X side must not be squished by the titlebar height.
+    let data = f.testwl.get_surface_data(id).unwrap();
+    let viewport = data.viewport.as_ref().unwrap();
+    assert_eq!(
+        (viewport.width, viewport.height),
+        (200, 200),
+        "viewport size"
+    );
+
```
En-En-Code: Cut--the previous asserts already cover this behavior (this portion of the test is checking the titlebar expands with the window).

En-En-Code (changes requested):
Cool. Mostly just some test clean up.

Fixes #472, #501, #504.

src/server/event.rs
mdsitton: Technically on this commit no. I do have 2 branches with further changes that i haven't made pr's for yet that made use of it but fair enough about having it as a more general refactor PR.

mdsitton:
Thanks for the review, working on the fixes now.

src/server/tests.rs
mdsitton: should be resolved with the latest commit

src/server/tests.rs
mdsitton: Should also be fixed now

src/server/tests.rs
mdsitton: Done

src/server/tests.rs
mdsitton: also done

src/server/tests.rs
```diff
@@ -3149,42 +3149,51 @@ fn client_side_decorations() {
...
-        f.run();
-
-        f.testwl
-            .force_decoration_mode(id, zxdg_toplevel_decoration_v1::Mode::ClientSide);
         f.testwl.configure_toplevel(id, 100, 100, vec![]);
         f.run();
         f.run();
 
+        let x_width = (100.0 * scale).floor() as u16;
+        let x_height = (content_height * scale).floor() as u16;
```
En-En-Code, line 3185: [Float-to-int casts round to zero](https://doc.rust-lang.org/reference/expressions/operator-expr.html#r-expr.as.numeric.float-as-int) already.

src/server/tests.rs
```diff
@@ -3149,42 +3149,51 @@ fn client_side_decorations() {
...
             "X window size at scale {scale}"
         );
         let data = f.testwl.get_surface_data(id).unwrap();
         let viewport = data.viewport.as_ref().unwrap();
         assert_eq!(
             (viewport.width, viewport.height),
-            (100, 75),
+            (100, content_height as i32),
             "viewport size at scale {scale}"
         );
```
En-En-Code, line 3198: DRY; use the helper function you have written for this.

En-En-Code (changes requested):
Just a couple more test nits.

src/server/tests.rs
```diff
@@ -3129,14 +3142,63 @@ fn client_side_decorations() {
...
     assert!(f.testwl.get_surface_data(subsurface_id).is_none());
     assert!(!subsurface.is_alive());
 }
 
+#[test]
+fn client_side_decorations_fractional_scale() {
+    use super::decoration::DecorationsDataSatellite;
+
+    // The titlebar is subtracted in logical pixels, so the X window content and the viewport
+    // agree at fractional scales as well.
```
Supreeeme:
```suggestion
```

Supreeeme:
Fixed some stuff and merged it in [5274bdc](https://github.com/Supreeeme/xwayland-satellite/commit/5274bdce3dd96c922c73b00276b2a628356e7f2b), thanks.

PR #497, opened 2026-09-07
feat: set _NET_CLIENT_LIST

etrobert:
EWMH section 3.2 says `_NET_CLIENT_LIST` SHOULD be set and updated by the
window manager. The atom is already interned, but never written, and
`_NET_SUPPORTED` doesn't advertise it.

Dofus 3 (Unity) needs it: its display plugin scans the list for a window
whose `_NET_WM_PID` matches its own, and enumerates monitors from that.
With no list it finds no window, ends up with zero displays, and aborts
before rendering anything — the game is just a black window under
satellite. Publishing the list is enough to get it running.

Windows are tracked in mapping order as the spec requires, and override
redirect windows are excluded since the list is scoped to managed windows.


nineluj:
Thanks @etrobert , the patch also worked for me on Dofus 3. I'm running NixOS + niri.

